PR #7, opened 2022-07-20
Tailwind transforms

AlanBreck:
This PR will add transforms for exporting our design system to a Tailwind preset which can be easily included in any Tailwind projects.

- adds tailwind transformations
- adds fonts for tailwind config
- adds shadows, gradients, and radii to tokens
- moves variables to preflight plugin
- adds font preflight plugin


tokens/config.js
```diff
@@ -1,6 +1,60 @@
...
+                ...subitems[0],
+                extensions: itemValue.extensions
+              };
+
+              contents[category][type][item].value = subitems.filter(v => v && v.value).map(v => v.value);
+            }
+          }
+        }
+      }
+      return contents;
```
petemill, line 27: Perhaps a little comment about what this parser does, since it doesn't have a function or file name?

petemill (approved):
Just left 1 comment about a comment!

tokens/config.js
AlanBreck: Good point. In addition, should this also be pulled into its own file? I debated on that.

PR #10, opened 2022-07-29
Legacy colors

petemill:
Adds colors from the pre-leo design system, since apparently they're still being used in new figma designs. Developers have an active need to use these variables, so we'll include them. But, to make sure developers don't use them just because the colors match, or at least know they are not intended to be used forever, we'll use the `"legacy"` prefix.

Output example:

```css
  --color-legacy-text1: rgb(33, 37, 41);
  --color-legacy-light-mode-text1: rgb(33, 37, 41);
  --color-legacy-dark-mode-text1: rgb(0, 0, 0);
@media (prefers-color-scheme: light) {
  --color-legacy-text1: rgb(0, 0, 0);
}
```

### DIFF
--- a/tokens/config.js
+++ b/tokens/config.js
@@ -6,11 +6,11 @@ module.exports = {
       buildPath: "build/css/",
       files: [
         {
-          destination: "_variables.css",
+          destination: "variables.css",
           format: "custom/css",
           filter: "filterWeb",
           options: {
-            showFileHeader: false,
+            showFileHeader: true,
           },
         },
       ],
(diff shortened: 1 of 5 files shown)


tokens/config.js
```diff
@@ -6,11 +6,11 @@ module.exports = {
       buildPath: "build/css/",
       files: [
         {
-          destination: "_variables.css",
+          destination: "variables.css",
```
AlanBreck, line 9: Nit: I kinda like the `_` prefix as it hints that it's a partial to be included with other things. I'm not married to it, but just curious about your thoughts for dropping it.

AlanBreck (approved):

tokens/config.js
petemill: My thought was it's a bit weird for the main import for dependent projects to be prefixed with an _ which was seeming like a private internal file.

PR #12, opened 2022-08-07
css variables: additional variables for colors omitting the rgb() function call

petemill:
So that colors may be used with custom opacity at the point of use. The new variables have the suffix "-rgb"

### DIFF
--- a/tokens/transformation/web/formatCss.js
+++ b/tokens/transformation/web/formatCss.js
@@ -1,50 +1,75 @@
 const formattedVariables = require('./formattedVariables')
 const fileHeader = require('./fileHeader')
 
-const filteredTokens = (dictionary, filterFn) => {
+function isColor(token) {
+  return token.type === 'color'
+}
+
+function duplicateTokenWithRgbText(token) {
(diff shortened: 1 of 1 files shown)


tokens/transformation/web/formatCss.js
```diff
@@ -1,50 +1,75 @@
...
+function isColor(token) {
+  return token.type === 'color'
+}
+
+function duplicateTokenWithRgbText(token) {
+  const rgbStart = 'rgb('
+  const rgbEnd = ')'
+  if (token.value && token.value.startsWith(rgbStart) && token.value.endsWith(rgbEnd)) {
+    const newToken = { ...token }
+    newToken.value = token.value.slice(rgbStart.length).slice(0, rgbEnd.length * -1)
```
petemill, line 13: Not sure why I've used 2 slices here!

petemill:
Evaluate these as part of consideration for including this:
- https://github.com/brave-experiments/leo/issues/17 - use tailwind theme function and postcss instead
- Is adding more css variables a performance issue https://github.com/brave-experiments/leo/issues/15

PR #19, opened 2022-08-11
adds build.yml workflow

AlanBreck:
Adds basic deploy script so that we can implement a super simple check for PRs to ensure that build works.


fallaciousreasoning (approved):
LGTM! I think we should try and get this in

PR #20, opened 2022-08-22
Added Skia color config

nullhook:
This PR generate output for Skia colors in a single header file.

1. Removes "LightMode" and "DarkMode" from names
2. adds two specific namespaces

### DIFF
--- a/tokens/transformTokens.js
+++ b/tokens/transformTokens.js
@@ -1,9 +1,10 @@
 const StyleDictionary = require('style-dictionary')
 const baseConfig = require('./config')
 
-// Register Web transforms
+// Register transforms
 require('./transformation/web')
 require('./transformation/tailwind')
+require('./transformation/skia')
 
 // Register Global transforms
 StyleDictionary.registerTransform({
--- a/tokens/transformation/skia/index.js
+++ b/tokens/transformation/skia/index.js
@@ -1,0 +1,55 @@
+const StyleDictionary = require('style-dictionary')
+const _template = require('lodash/template')
+const fs = require('fs')
+
+StyleDictionary.registerTransformGroup({
+  name: 'skia',
+  transforms: StyleDictionary.transformGroup.css.concat([
+    'color/hex8ToRgbaPartial'
+  ])
+})
+
+const filteredTokens = (dictionary, filterFn) => {
+  let filtered = dictionary.allTokens;
+  if (typeof filterFn === "function") {
+    filtered = dictionary.allTokens.filter(token => filterFn(token))
+  }
+
+  filtered = filtered.map(tokens => {
+    // convert to hex
+    const value = tokens.value.split(',')
+    .map(s => '0x' + Number(s).toString(16).padStart(2, '0').toUpperCase())
+    // convert to chromium's pascalCase convention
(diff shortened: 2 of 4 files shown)


tokens/transformTokens.js
```diff
@@ -4,6 +4,7 @@ const baseConfig = require('./config')
 // Register Web transforms
 require('./transformation/web')
 require('./transformation/tailwind')
+require('./transformation/skia')
```
AlanBreck, line 7: Nit: Since this isn't a web transform, maybe we should rename the comment above to "Register transforms", or something.

tokens/transformation/skia/index.js
```diff
@@ -0,0 +1,61 @@
+const StyleDictionary = require('style-dictionary')
+const _template = require('lodash/template')
+const fs = require('fs')
+
+StyleDictionary.registerTransform({
+  name: 'color/hex8ToRgbaPartial',
+  ...require('../common/colorToRgbaPartial')
+})
```
AlanBreck, line 8: This shouldn't be necessary since it's already registered by both the web and trailwind transformations. Both of those should probably be eventually refactored as well. We should probably consider importing common transforms explicitly above the other imports in `transformTokens.js`.

tokens/transformation/skia/index.js
```diff
@@ -0,0 +1,61 @@
...
+  let filtered = dictionary.allTokens;
+  if (typeof filterFn === "function") {
+    filtered = dictionary.allTokens.filter(token => filterFn(token))
+  }
+
+  filtered = filtered.map(tokens => {
+    // convert to hex
+    const value = tokens.value.split(',')
+    .map(s => '0x' + Number(s).toString(16).padStart(2, '0').toUpperCase())
+    // convert to chromium's camelCase convention
```
AlanBreck, line 28: nit: would this be pascal case?

tokens/transformation/skia/templates/colors.h.template
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2022 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// you can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "third_party/skia/include/core/SkColor.h"
+
+namespace leo::light {
+  <%= groupedTokens.light.allTokens.map(prop => {
+    return `constexpr SkColor ${prop.name} = SkColorRGB(${prop.value})`;
```
sangwoo108, line 10: Hi! With all due respect, if this line is for c++ header, should we use `SkColorSetRGB()` instead of `SkColorRGB()`?

Roughly, 
```js
    return `constexpr SkColor ${prop.name} = SkColorSetRGB(${prop.value[0]}, ${prop.value[1]}, ${prop.value[2]})`;
```

nullhook: makes sense

tokens/transformation/skia/index.js
nullhook: left over, good catch

AlanBreck (approved):

PR #21, opened 2022-08-24
Big update on colors and font values

aguscruiz:
Updated font values and color values.
Renamed secondary to primary and viceversa. Also added a "primary-legacy", which is the blurple. New color is more blue-tinted.

Removed a lot of unused gradients and updated elevation values

### DIFF
--- a/web-components/button/button.svelte
+++ b/web-components/button/button.svelte
@@ -80,27 +80,28 @@
 
   // Kind Variations
   .leoButton.isPrimary {
-    --bg: var(--color-secondary-light-mode-70);
-    --bg-hover: var(--color-secondary-light-mode-80);
-    --bg-active: var(--color-secondary-light-mode-90);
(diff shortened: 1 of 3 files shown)


AlanBreck (approved):
I've just updated the `button` component to reflect the primary/secondary color change.

web-components/button/button.svelte
```diff
@@ -80,48 +80,50 @@
...
     --color-loading: var(--color-gray-70);
     --border-width: 1px;
     --border-color: var(--color-gray-30);
     &[disabled] {
       opacity: 0.5;
     }
     @media (prefers-color-scheme: dark) {
-      --bg: black;
-      --color-hover: var(--color-secondary-dark-mode-80);
+      --bg: transparent;
```
petemill, line 112: we can remove this line, I believe

petemill (approved):

PR #25, opened 2022-09-01
PostCSS plugin for @darktheme directive

fallaciousreasoning:
Hey Pete, what do you think of this? I had a bit of trouble debugging and getting it to work nicely with Svelte, but I think it's working now.

### DIFF
--- a/postcss/darkmode.js
+++ b/postcss/darkmode.js
@@ -13,7 +13,7 @@ ${indent(rule.nodes.map(n => n.toString() + ';').join('\n'), 1)}
 }`
 }
 
-module.exports = (options = { forceGlobal: true }) => {
+module.exports = (options = { forceGlobal: false }) => {
     return {
         postcssPlugin: 'darkmode',
         AtRule: {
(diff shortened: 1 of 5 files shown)


postcss/darkmode.js
```diff
@@ -0,0 +1,40 @@
...
+        console.log(rule.toString())
+        return ''
+    }
+    const selectors = rule.selector.split(',').map(s => s.trim()).map(transform)
+    return `${selectors.join(', ')} {
+${indent(rule.nodes.map(n => n.toString() + ';').join('\n'), 1)}
+}`
+}
+
+module.exports = (options = { forceGlobal: true }) => {
```
AlanBreck, line 16: Can we actually default this to false? It seems the most intuitive would be to assume it won't, and then change that flag explicitly in the svelte config.

AlanBreck (commented):
This is awesome!

postcss/darkmode.js
fallaciousreasoning: Yup, I went back and forwards on this a few times but happy to default it to false :smile:

AlanBreck:
Should we close this since you've moved to a different repo?

fallaciousreasoning:
Yeah, probably a good call. Closing (see https://github.com/fallaciousreasoning/postcss-plugin-darkmode)

PR #26, opened 2022-09-06
Autogenerate Type Definitions for Svelte Components

fallaciousreasoning:
At the moment, these Type Definitions are dumped into the `types` folder, and you need to run
`npx rollup -c rollup.config.js` or `npx rollup -c rollup.config.js --watch` to regenerate the type defintions.

In the future, I would like to do the following:
1. Automatically generate the `react.ts` wrapper for each component. I have a working prototype, but it doesn't play too nicely with the `example-ui-react` project (yet!).
2. Not have the `types` be part of the repository (maybe generate at install time?).

What are your thoughts on this @AlanBreck & @petemill?


petemill (commented):
Leveraging svelte2tsx is a great approach. I'm wondering a couple of things though
- Does using rollup make it a lot easier? It doesn't seem too bad as I don't think it's enforcing consuming projects to use rollup, Leo would simply be using it for internal 'compile' output. But it does add a little complexity. Possible it's better than finding all *.svelte files and piping them to svelte2tsx?
- Do we need to generate .d.ts files for `svelte-react.ts`? It's a file we control and doesn't seem to differ dynamically based on any svelte files. Same question for `[component]/react.ts`.
- With regards to putting the .d.ts files in a `types` directory, my opinion could go either way. If we can reduce the need to a single `[component].svelte..d.ts` file then perhaps inline is more clear as to where types are coming from. However, another directory could be more clear that the files are generated and should not be modified by the dev.
- Seems like we should have a package.json command, something like `npm run gen-web-component-types`.

fallaciousreasoning:
As per your suggestion, I've done the following:
- [x] Don't use rollup - we have a custom script that does this now.
- [x] Only include the `*.d.ts` files for the `.svelte` components themselves.
- [x] Put the generated `*.d.ts` files next to their relevant svelte component.
- [x] Add a build script for doing this. 

In a follow up PR, I've done something similar with the `react` bindings (see #44).

I wasn't sure whether we should include the generated `.d.ts` file in the repository, or if it should be `.gitignore`'d and generated on build/install/publish/whatever. Interested to here your thoughts here @petemill @AlanBreck

petemill (approved):

package.json
```diff
@@ -36,14 +33,19 @@
...
     "jest": "^27.0.6",
+    "postcss": "^8.4.14",
+    "postcss-js": "^4.0.0",
     "sass": "^1.53.0",
     "sass-loader": "^13.0.2",
     "style-loader": "^3.3.1",
-    "svelte-loader": "^3.1.3"
+    "svelte-loader": "^3.1.3",
+    "svelte2tsx": "^0.5.19",
+    "typescript": "^4.8.4"
```
petemill, line 43: can we move this to a peer dependency

PR #27, opened 2022-09-07
[Component]: Collapse

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/53

This adds a simple collapse component to Leo based on the UX mockup in [Figma](https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?node-id=278%3A12772)

Storybook:
https://leo-storybook-git-wc-collapse-fallaciousreasoning.vercel.app/

### DIFF
--- a/web-components/collapse/collapse.stories.svelte
+++ b/web-components/collapse/collapse.stories.svelte
@@ -1,0 +1,40 @@
+<script>
+  import { Meta, Story, Template } from '@storybook/addon-svelte-csf'
+
+  import Collapse from './collapse.svelte'
+</script>
+
+<Meta
+  title="Collapse"
+  component={Collapse}
+  argTypes={{
+    open: { control: 'select', options: [undefined, true, false] },
+  }}
+/>
+
+<Template let:args>
+  <Collapse {...args} />
+</Template>
+
+<Story name="Primary" let:args>
+  <Collapse title="Title" {...args}>
+    <!-- TODO(fallaciousreasoning): When we get icons, use one of those here (see https://github.com/brave/leo/issues/45). -->
+    <svg
+      slot="icon"
+      viewBox="0 0 24 24"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        fill-rule="evenodd"
+        clip-rule="evenodd"
+        d="M11.9965 22C14.6691 22 17.1807 20.96 19.0692 19.0717C20.9593 17.1825 22 14.6708 22 12C22 9.32917 20.9593 6.8175 19.0692 4.92917C17.1807 3.04 14.6691 2 11.9965 2C9.32481 2 6.81326 3.04 4.92393 4.92917C1.02536 8.8275 1.02536 15.1717 4.92393 19.0708C6.81326 20.96 9.32481 22 11.9965 22ZM11.9965 3.66667C14.2238 3.66667 16.3169 4.53333 17.8909 6.1075C19.4658 7.68167 20.3329 9.77417 20.3329 12C20.3329 14.2258 19.4658 16.3192 17.8909 17.8925C16.3169 19.4667 14.2238 20.3333 11.9965 20.3333C9.77011 20.3333 7.67701 19.4667 6.10299 17.8925C2.85348 14.6433 2.85348 9.35667 6.10299 6.1075C7.67701 4.53333 9.77011 3.66667 11.9965 3.66667ZM10.3303 7.8325C10.3303 8.75167 11.0783 9.5 11.9974 9.5C12.9173 9.5 13.6645 8.75167 13.6645 7.8325C13.6645 6.91417 12.9173 6.16667 11.9974 6.16667C11.0783 6.16667 10.3303 6.91417 10.3303 7.8325ZM12 11.2C11.5306 11.2 11.15 11.5806 11.15 12.05V17.15C11.15 17.6194 11.5306 18 12 18C12.4695 18 12.85 17.6194 12.85 17.15V12.05C12.85 11.5806 12.4695 11.2 12 11.2Z"
+        fill="currentColor"
+      />
+    </svg>
+    Content goes here
+  </Collapse>
+</Story>
+
+<style>
+</style>
--- a/web-components/collapse/collapse.svelte
+++ b/web-components/collapse/collapse.svelte
@@ -1,0 +1,134 @@
+<svelte:options tag="leo-collapse" />
+
+<script lang="ts">
+  import { createEventDispatcher } from 'svelte'
+
+  export let title = ''
+
+  // We have a separate external and internal isOpen flag because if the parent
+  // doesn't pass anything in we want to be able to toggle open & closed
+  // normally. However, if they didn't pass anything in, we just emit an event
+  // and let the parent handle toggling the flag.
+  export let isOpen: boolean | undefined
+  $: isOpenInternal = isOpen ?? false
+
+  const dispatcher = createEventDispatcher()
+  const toggle = (e) => {
+    e.preventDefault()
+
+    // We only support toggling if this component isn't being 'controlled'
+    const toggleTo = !isOpenInternal
+    if (isOpen === undefined) isOpenInternal = toggleTo
+    dispatcher('toggle', { open: toggleTo })
+  }
+</script>
+
+<details class="leoCollapse" open={isOpenInternal}>
+  <summary on:click={toggle}>
+    {#if $$slots.icon}
+      <div class="icon">
+        <slot name="icon" />
+      </div>
+    {/if}
+    <div class="title">
+      {title}
+    </div>
+    <div class="arrow">
+      <svg
+        width="16"
+        height="9"
+        viewBox="0 0 16 9"
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
+      >
+        <path
+          fill-rule="evenodd"
+          clip-rule="evenodd"
+          d="M0.234303 0.234362C0.546723 -0.0780572 1.05325 -0.0780572 1.36567 0.234362L7.79999 6.66868L14.2343 0.234363C14.5467 -0.078056 15.0533 -0.0780559 15.3657 0.234363C15.6781 0.546783 15.6781 1.05331 15.3657 1.36573L8.64852 8.08289C8.17989 8.55152 7.42009 8.55152 6.95146 8.08289L0.234303 1.36573C-0.0781163 1.05331 -0.0781163 0.546782 0.234303 0.234362Z"
+          fill="currentColor"
+        />
+      </svg>
+    </div>
+  </summary>
+  <div class="content">
+    <slot />
+  </div>
+</details>
+
+<style lang="scss">
+  .leoCollapse {
+    --background-color: var(--color-white);
+    --leo-icon-size: var(--icon-size, 24px);
+    --animation-duration: 0.12s;
+
+    @theme (dark) {
+      --background-color: var(--color-gray-10);
+    }
+
+    background-color: var(--background-color);
+    box-shadow: var(--effect-elevation-01);
+    border-radius: var(--radius-8);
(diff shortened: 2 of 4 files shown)


web-components/collapse/collapse.svelte
```diff
@@ -0,0 +1,131 @@
...
+        --background-color: var(--color-primary-10);
+      }
+    }
+
+    &:hover .title {
+      color: var(--color-primary-80);
+    }
+
+    &:focus-within:has(summary:focus-visible) {
+      box-shadow: 0px 0px 0px 1.7px rgba(255, 255, 255, 0.5),
```
petemill, line 88: This should be a token, but we need this first
https://github.com/brave-experiments/leo/issues/35

web-components/collapse/collapse.svelte
```diff
@@ -0,0 +1,131 @@
+<svelte:options tag="leo-collapse"/>
+
+<script lang="ts">
+import { createEventDispatcher } from "svelte";
+
+
+  export let title = ''
+  export let open: boolean | undefined
```
petemill, line 8: nit: reduce to 1 line so we can easily see default value for attribute:
```
export let open: boolean | undefined = false
```

What do we think about naming boolean variables `isXYZ`. Usually I'd suggest that for React, but since this is more "native html-like", I'm unsure. But still, svelte does accept complex data for things, so sometimes it's difficult to tell if an attribute expects data or a simple boolean. In React, the main reason is that `open` could be a function to handle open instead of a directive (or reflection) to be open.

Either way, I'd like to be consistent accross components so we should change this or the button component and move forward with that.

@fallaciousreasoning @AlanBreck @nullhook

web-components/collapse/collapse.svelte
petemill: Also would be nice to have a transition on the shadow, similar to button. So we can have focus styles all transition consitently between elements.

petemill (changes requested):
This is great, but seems the hover colors are red when they should in fact be purple. I tried rebasing on main and it fixed it so I think you just need a rebase. Then just some other minor comments, especially being able to customize the icon.

web-components/collapse/collapse.svelte
AlanBreck: I personally like the `isXYZ` naming convention.

web-components/collapse/collapse.svelte
```diff
@@ -0,0 +1,131 @@
...
+    // We only support toggling if this component isn't being 'controlled'
+    const toggleTo = !isOpen;
+    if (open === undefined) isOpen = toggleTo
+    dispatcher('toggle', { open: toggleTo })
+  }
+</script>
+
+<details class="leoCollapse" open={isOpen}>
+  <summary on:click={toggle}>
+    <svg
```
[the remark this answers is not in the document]
fallaciousreasoning: Wasn't sure if the icon was mean to be configurable :smile: Updated & added a css variable for sizing.

web-components/collapse/collapse.svelte
fallaciousreasoning: I've updated the name of the prop to be `isOpen`. The reason the props are split like this is so we can support two different scenarios:
1) Uncontrolled: The component Just Works when you drop it in, it can be expanded and collapsed.
2) Controlled: The owner of the component passes in `isOpen` and the component fires an event when it thinks it should be expanded/collapsed. An example of when this is useful might be when you want to be able to toggle the collapse with an external button.

Maybe there's a more Sveltey way of doing this, but this way ensures we get nice interop with React.

I've renamed the internal variable to be `isOpenInternal` to make this more obvious (we need to know that `isOpen` is undefined in order to control the collapse).

web-components/collapse/collapse.svelte
fallaciousreasoning: Done :+1:

fallaciousreasoning:
Rebased (colors are now blue!), added the slot, focus transitions & renamed `open` to `isOpen`

web-components/collapse/collapse.svelte
```diff
@@ -0,0 +1,131 @@
+<svelte:options tag="leo-collapse" />
+
+<script lang="ts">
+  import { createEventDispatcher } from 'svelte'
+
+  export let title = ''
+  export let isOpen: boolean | undefined
+  $: isOpenInternal = isOpen ?? false
```
AlanBreck, line 8: Why do we need two variables here? I believe this could be changed to

```suggestion
  export let isOpen: boolean | undefined = false;
```

web-components/collapse/collapse.svelte
```diff
@@ -0,0 +1,131 @@
...
+    if (isOpen === undefined) isOpenInternal = toggleTo
+    dispatcher('toggle', { open: toggleTo })
+  }
+</script>
+
+<details class="leoCollapse" open={isOpenInternal}>
+  <summary on:click={toggle}>
+    {#if $$slots.icon}
+      <div class="icon">
+        <slot name="icon" />
```
AlanBreck, line 25: Do we want a default icon here?

web-components/collapse/collapse.svelte
```diff
@@ -0,0 +1,131 @@
...
+      flex-direction: row;
+      align-items: center;
+      gap: var(--spacing-8);
+
+      padding: 16px 18px;
+
+      list-style: none;
+      cursor: pointer;
+    }
+  }
```
AlanBreck, line 100: Nit: I would be in favor of de-nesting here.

web-components/collapse/collapse.stories.svelte
```diff
@@ -0,0 +1,39 @@
...
+    >
+      <path
+        fill-rule="evenodd"
+        clip-rule="evenodd"
+        d="M11.9965 22C14.6691 22 17.1807 20.96 19.0692 19.0717C20.9593 17.1825 22 14.6708 22 12C22 9.32917 20.9593 6.8175 19.0692 4.92917C17.1807 3.04 14.6691 2 11.9965 2C9.32481 2 6.81326 3.04 4.92393 4.92917C1.02536 8.8275 1.02536 15.1717 4.92393 19.0708C6.81326 20.96 9.32481 22 11.9965 22ZM11.9965 3.66667C14.2238 3.66667 16.3169 4.53333 17.8909 6.1075C19.4658 7.68167 20.3329 9.77417 20.3329 12C20.3329 14.2258 19.4658 16.3192 17.8909 17.8925C16.3169 19.4667 14.2238 20.3333 11.9965 20.3333C9.77011 20.3333 7.67701 19.4667 6.10299 17.8925C2.85348 14.6433 2.85348 9.35667 6.10299 6.1075C7.67701 4.53333 9.77011 3.66667 11.9965 3.66667ZM10.3303 7.8325C10.3303 8.75167 11.0783 9.5 11.9974 9.5C12.9173 9.5 13.6645 8.75167 13.6645 7.8325C13.6645 6.91417 12.9173 6.16667 11.9974 6.16667C11.0783 6.16667 10.3303 6.91417 10.3303 7.8325ZM12 11.2C11.5306 11.2 11.15 11.5806 11.15 12.05V17.15C11.15 17.6194 11.5306 18 12 18C12.4695 18 12.85 17.6194 12.85 17.15V12.05C12.85 11.5806 12.4695 11.2 12 11.2Z"
+        fill="currentColor"
+      />
+    </svg>
+    Content goes here</Collapse
+  >
```
AlanBreck, line 35: Uber nit (I think prettier botched this):

```suggestion
    Content goes here
  </Collapse>
```

web-components/collapse/collapse.svelte
fallaciousreasoning: Ah yeah, Pete asked this as well. It's so we can have a controlled and uncontrolled state for the collapse. There might be a better way of doing this in Svelte but basically, I wanted it to work normally without specifying `isOpen` on the component (it should toggle itself) and also in the React way, where you set `isOpen` and it doesn't toggle when you click it, it just emits an event.

web-components/collapse/collapse.svelte
fallaciousreasoning: I was thinking it's probably fine to have no icon by default, but maybe we want to ensure all collapses have a default icon, not sure.

web-components/collapse/collapse.stories.svelte
fallaciousreasoning: Oh yeah, that's some odd formatting. I've noticed it struggles a bit more with Svelte files than with React :/

web-components/collapse/collapse.svelte
AlanBreck: I think I understand. In your example of a more idiomatic solution, I don't think `bind:` is necessary?

fallaciousreasoning: Yeah, so basically, I want to support both of these cases:

```svelte

<!-- Parent doesn't care -->
<Parent>
	<Collapse>Stuff & Things</Collapse>
</Parent>

<!-- Parent is in charge -->
<script>
     let open = false;
</script>
<button on:click={() => open = false}>Close the thing!</button>
<Collapse isOpen={open} on:toggle={newOpen => open = newOpen}>Maybe hidden?</Collapse>
```


<script>
    let open = false;
</script>
<button on:click={() => open = false}>Close the thing 2.0!</button>
<Collapse bind:isOpen={open}>Maybe hidden?</Collapse>

fallaciousreasoning:
@petemill mind taking another look?

aguscruiz (commented):
Looks great! Just a couple minor nitpicks:

- Default fonts inside the panel should be Poppins (This applies to everything, default font should always be Poppins)
- When inspecting, the icon color seems to be gray-120, but we don't have a 120 anymore, just bringing it up. The color should be the token `light-icon-default`. This is both for the left and the right icon. And every icon should default to being `light-icon-default unless the design says otherwise

fallaciousreasoning:
@aguscruiz mind taking another look at this?

web-components/collapse/collapse.svelte
```diff
@@ -0,0 +1,178 @@
...
+    );
+    --shadow: var(--leo-collapse-shadow, var(--effect-elevation-01));
+    --shadow-hover: var(
+      --leo-collapse-shadow-hover,
+      var(--effect-elevation-02)
+    );
+    --shadow-focus: var(
+      --leo-collapse-shadow-focus,
+      0px 0px 0px 1.5px rgba(255, 255, 255, 0.5),
+      0px 0px 4px 2px #423eee
```
AlanBreck, line 86: Is this not already in a token?

fallaciousreasoning: Ah, I didn't find it because it wasn't named in the figma file, and the focus state didn't match the one in variables.css. Updated it to use `--effect-focus-state`.

web-components/collapse/collapse.svelte
```diff
@@ -0,0 +1,181 @@
...
+
+    @media (prefers-reduced-motion) {
+      --transition-duration: 0s;
+    }
+
+    background-color: var(--background-color);
+    box-shadow: var(--shadow);
+    border-radius: var(--radius);
+    border: 1px solid var(--border-color);
+    transition: box-shadow var(--transition-duration) ease-in-out,
```
AlanBreck, line 107: @fallaciousreasoning, can we include `color` in the transition so that the title transitions along with the background on hover?

AlanBreck (approved):

aguscruiz (approved):
Everything looks great! Great work as always @fallaciousreasoning ✨

web-components/collapse/collapse.svelte
fallaciousreasoning: :+1: good call. Turns out I'd tried to do this but written `--var(--transition-duration)` which, believe it or not, is not how you're meant to do it :laughing:

AlanBreck: It's so buttery smooth! Thank you!

PR #29, opened 2022-09-12
Add postcss @darkmode plugin

fallaciousreasoning:
The darkmode plugin is here: https://github.com/fallaciousreasoning/postcss-plugin-darkmode. Not sure if this should be moved to the Brave or Brave Experiments organisation.


AlanBreck (approved):

fallaciousreasoning:
After having a chat with Pete, we're going to make the following changes before landing this:
- [ ] Change the `AtRule` `@theme (dark)`
- [ ] Add support for `@theme (light)`
- [ ] Move the code from the external package into the repository.

Separately, we need to come up with a way for doing this from `styled-components`

fallaciousreasoning:
@AlanBreck @petemill mind taking another look. I've moved the plugin into this repo, and added support for `@theme (dark|light)`

AlanBreck:
This looks so good! One tiny nit I was going to ask about is removing the parens around the theme name (e.g. `@theme light`), but it appears it works with or without. Yay!

postcss/theme.js
```diff
@@ -0,0 +1,180 @@
...
+        .replace(regex, '\\$1')
+}
+
+const splitRule = (rule, selectorToExtract) => {
+    // |cloneAfter|, so in an |each| loop, the new rule will be processed.
+    const cloned = rule.cloneAfter();
+
+    // Remove the light selector from the cloned rule.
+    // Note: Directly splicing |selectors| has no effect, we have
+    // to set a varible.
```
AlanBreck, line 17:
```suggestion
    // to set a variable.
```

postcss/theme.js
```diff
@@ -0,0 +1,180 @@
...
+
+        // Remove all of the overridden light rules.
+        for (const decl of Object.values(properties)) {
+            if ('remove' in decl.base)
+                nodesToDelete.add(decl.base);
+        }
+
+        return {
+            dark: darkVariables,
+          	light: lightVariables,
```
AlanBreck, line 128:
```suggestion
            dark: darkVariables,
            light: lightVariables,
```

postcss/theme.js
```diff
@@ -0,0 +1,180 @@
...
+    return `--${selector}_${decl.prop}`
+        .replace(/\s+/g, '_')
+        .replace(regex, '\\$1')
+}
+
+const splitRule = (rule, selectorToExtract) => {
+    // |cloneAfter|, so in an |each| loop, the new rule will be processed.
+    const cloned = rule.cloneAfter();
+
+    // Remove the light selector from the cloned rule.
```
AlanBreck, line 15: This is actually removing the base selector, not necessarily `light`, right?

postcss/theme.js
```diff
@@ -0,0 +1,180 @@
...
+    // to set a varible.
+    const index = cloned.selectors.indexOf(selectorToExtract);
+    const copy = [...cloned.selectors];
+    copy.splice(index, 1);
+    cloned.selectors = copy;
+
+    // If we don't set this all the formatting breaks.
+    cloned.raws.before = '\n';
+
+    // Set the selectors on the light rule to be just the light selector.
```
AlanBreck, line 26: Same question about this comment.

postcss/theme.js
```diff
@@ -0,0 +1,180 @@
...
+            for (const baseSelector of [...rule.selectors]) {
+                if (!selectors.has(baseSelector)) continue
+
+                // If the rule has multiple selectors, we need to split the the
+                // rule apart:
+                // For example, when processing the rule .frob:
+                // .foo, .frob, .bar { background: red; }
+                // should become
+                // .frob { background: red; }
+                // .foo, .bar { background: red; }
```
AlanBreck, line 57: Why would it become 2 rules instead of three, given that there are three selectors?

AlanBreck (commented):
One thing to note is that we almost certainly don't want these generated variables to appear in intellisense. Depending on how we set up the build process, and how CSS variable intellisense plugins work, we don't want them indexing anything found in our Svelte components.

postcss/theme.js
fallaciousreasoning: Because at this point, we only know that `.frob` is themed (it has a light/dark variation), so we might as well leave `.foo, .bar` as a combined selector rather than duplicating it and making a larger CSS file. If we encounter a theme variation for either of them later we can split them then.

This function splits the themed selector out from the other ones.

postcss/theme.js
fallaciousreasoning: Nice catch! I missed that when adding light theme support :laughing:

fallaciousreasoning:
I've fixed the nits @AlanBreck. Yeah, I went with the parens to make it a bit more similar to `(prefers-color-scheme: dark)` but they are optional :)

WRT the crazy CSS variables showing in Intellisense, I'm not 100% sure what to do. I'm pretty sure they won't show up for Svelte components, and looking at the generated output for webcomponents, I don't think they'll show there either (as the style is basically set with a 
```js
shadowRoot.innerHTML = `<style>....</style>`
```
so maybe we're okay here?

AlanBreck:
> WRT the crazy CSS variables showing in Intellisense, I'm not 100% sure what to do

Me either. I'm good with crossing this bridge only if/when we have to.

PR #42, opened 2022-10-04
Improve Button type definitions

fallaciousreasoning:
This does the following:
1) Makes it possible to passing through button attributes.
2) Updates the type definitions such that the button is either a button or a link.
3) Adds the type definitions for the Button/A tag as optional props.

It does not:
1) Add support for events.

@petemill what do you think about this as an approach?

### DIFF
--- a/web-components/button/button.svelte
+++ b/web-components/button/button.svelte
@@ -4,11 +4,31 @@
   import { createEventDispatcher } from 'svelte';
   import type * as Props from './props'
 
+  type Href = $$Generic<string | undefined>;
+  type Disabled = $$Generic<Href extends string ? boolean : undefined>;
+    type ExcludedProps = 'size' | 'href' | 'hreflang';
+
+  interface CommonProps {
+    kind?: Props.ButtonKind;
+    size?: Props.ButtonSize;
+    isLoading?: boolean;
+  }
+
+  type ButtonProps = CommonProps & Omit<Partial<svelte.JSX.HTMLAttributes<HTMLElementTagNameMap['button']>>, ExcludedProps> & {
+    isDisabled?: Disabled;
+  }
+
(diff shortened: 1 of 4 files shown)


web-components/button/button.svelte
```diff
@@ -22,16 +42,36 @@
   }
 </script>
 
+<svelte:element
```
fallaciousreasoning, line 45: I have no idea why but the formatter in my VSCode really wants it to be this way. If you think the approach is reasonable I'll put this back to the way it was.

fallaciousreasoning:
Hey @AlanBreck do we want to try and land this?

AlanBreck:
Sounds good to me!!

fallaciousreasoning:
All right, I've fixed the merge conflict and the formatting. Anything else need doing?

AlanBreck:
Not that I can think of. The only thing that would be nice in the future is if there's any opportunity to simply for future component authors. No idea how, though. 😂

fallaciousreasoning:
Yeah, It's definitely pretty awkward at the moment - hopefully Svelte improves this :D

fallaciousreasoning:
@petemill what are your thoughts on this?

web-components/button/button.svelte
```diff
@@ -27,11 +61,12 @@
   .leoButton {
     // Gradients cannot have a transition, so we need to reset `transition`
     // to only apply to `box-shadow` and `border-color` in .isCTA
-    --default-transition: box-shadow .12s ease-in-out, color .12s ease-in-out, border-color .12s ease-in-out;
+    --default-transition: box-shadow 0.12s ease-in-out, color 0.12s ease-in-out,
+      border-color 0.12s ease-in-out;
```
AlanBreck, line 65: nit: don't love this formatting, but I'm assuming it's Prettier being opinionated?

AlanBreck (approved):

web-components/button/button.svelte
fallaciousreasoning: TBH, I don't even know if I'm using prettier. I'll revert this.

PR #75, opened 2022-11-06
Misc updated to tokens (elevation light-dark modes, etc.)

aguscruiz:
Various adjustments and improvements on tokens.


aguscruiz:
@AlanBreck this PR has the latest

fallaciousreasoning:
Do we want to land this?

aguscruiz:
I'd say yes

fallaciousreasoning:
All right, looks like there's a conflict. Shall we update the tokens to what's in Figma currently and land it?

AlanBreck:
Sounds good to me.

@aguscruiz, does this PR reflect the latest?

aguscruiz:
@AlanBreck yep!

AlanBreck:
Just scanning through the diff a bit, I believe this will result in quite a few breaking changes.

fallaciousreasoning:
Okay to fix the conflicts, and ensure everything is update to date, I think we should reset this PR to the latest main and remake the changes to `design-tokens.tokens.json`.

(in the leo directory, on this branch)
1. `git fetch`
2. `git reset --hard origin/main`
3. Update the design tokens again
4. Commit
5. Push

AlanBreck:
Yes, however this has potential to break things in existing projects like store, etc.

fallaciousreasoning:
Probably worth giving the components a once over too, to make sure nothing has broken.

AlanBreck:
Note that this also includes dark mode for elevations, so we'll have to update our transformers to include elevation in the light/dark media query logic.

aguscruiz:
Commenting on further updates I'm doing at the moment (not in this PR yet)

- Reorganization of less relevant font style tokens (i.e. Display tokens get put into the Components folder)
- Gradient / Hero updated
-

AlanBreck:
> Letter spacing for iOS and Android styles

From talking to Kyle, I don't think we can control letter spacing in iOS. Don't recall on Android.

AlanBreck:
I rebased the PR to resolve conflicts. It was a little weird because there were conflicts with `design-tokens.tokens.json`, which I wouldn't have anticipated since I'm surprised there would be any updates more recent than this PR. @aguscruiz, would you be able to spot check to ensure things look correct?

If it is correct (🤞), then the good news is that this PR is no where near as heavy as originally anticipated. There are some changes which will need to be made, though, especially for elevation. I'll try to get that done on Monday.

aguscruiz:
Looking at it seems correct, there are more changes that I haven't updated yet, so it's hard to say for sure.

.gitignore
```diff
@@ -7,3 +7,4 @@ node_modules/
 /svelte/
 /tokens/
 /icons-raw/
+/storybook-static/
```
fallaciousreasoning, line 10: Nit: no trailing newline :laughing:

fallaciousreasoning (approved):
This LGTM, but might be worth letting @petemill have a look as he's a bit more familiar with the tokens stuff

.gitignore
AlanBreck:
```suggestion
/storybook-static/

```

src/components/button/button.svelte
```diff
@@ -203,13 +203,44 @@
...
-    --bg-hover: var(--color-secondary-40);
-    --bg-active: var(--bg-hover);
+    --bg: transparent;
     --bg-focus: var(--bg);
     --bg-disabled: var(--color-gray-30);
     --color: white;
+    --default-bg-opacity: 1;
+
+    position: relative;
+    overflow: hidden;
```
petemill, line 215: This makes me a little nervous as it could cut off shadows, right? It doesn't seem to have any negative effect at the moment though so could be ok.

petemill (approved):

src/components/button/button.svelte
AlanBreck: Good call, and easily fixed. https://github.com/brave/leo/pull/75/commits/528337ca155d9ba5df6f1baa4bd984b047a576ad

PR #78, opened 2022-11-10
Theme plugin fixes

AlanBreck:
- fixes issues with postcss/theme plugin
- fixes hover color for button.isPrimary in dark mode

### DIFF
--- a/postcss/theme.js
+++ b/postcss/theme.js
@@ -8,6 +8,10 @@ const getPropertyName = (selector, decl) => {
         .replace(regex, '\\$1')
 }
 
+const wrapInGlobalSelector = (selector) => {
+    return `:global(${selector})`;
+}
+
 const splitRule = (rule, selectorToExtract) => {
     // |cloneAfter|, so in an |each| loop, the new rule will be processed.
     const cloned = rule.cloneAfter();
@@ -38,7 +42,7 @@ const defaultOptions = {
 module.exports = (options) => {
     options = { ...defaultOptions, ...options };
 
-    const rules = {};
+    let rules = {};
     const nodesToDelete = new Set();
     const findMatchingBaseRules = root => {
         const selectors = new Set(Object.keys(rules));
@@ -65,10 +69,10 @@ module.exports = (options) => {
 
     const extractDarkProperties = (selector, lightAndDark) => {
       	const variants = ['base', 'dark', 'light'];
-      
+
       	// At least one variant should have a value. Take it, and get the root node.
       	const root = variants.map(v => lightAndDark[v]).filter(v => v)[0].root();
-      
+
         // Selector is the same for light and for dark.
         const properties = {}
         const getPropertyVariants = (property) => {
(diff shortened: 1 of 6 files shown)


postcss/theme.js
```diff
@@ -38,7 +42,7 @@ const defaultOptions = {
 module.exports = (options) => {
     options = { ...defaultOptions, ...options };
 
-    const rules = {};
+    let rules = {};
```
fallaciousreasoning, line 45:
```suggestion
    const rules = {};
```

postcss/theme.js
```diff
@@ -136,6 +140,8 @@ module.exports = (options) => {
                 const theme = supportedThemes.find(t => atRule.params.includes(t));
                 if (!theme) throw new Error(`Encountered unsupported theme ${atRule.params}. Allowed themes are ${supportedThemes.join(', ')}`);
 
+                atRule.parent.hasThemeAtRule = true;
```
fallaciousreasoning, line 143: Is there a way we can set this on `root` instead? Because I worry that if the rule is deeply nested this means we won't run the plugin:

```scss
.foo {
     .bar {
          .baz {
                @theme (dark) {
                      background: red;
                 }
          }
     }
}
```

That said it might be fine, because I think only the compiled scss will get passed to the plugin.

postcss/theme.js
```diff
@@ -166,13 +182,18 @@ module.exports = (options) => {
...
+            }
 
             // Remove all empty nodes
             root.each(node => {
-                if (node.nodes.length === 0) node.remove();
+                if (!node.nodes || node.nodes.length === 0) node.remove();
             })
+
+            // Reset global vars
+            rules = {};
```
fallaciousreasoning, line 195: Does it maintain the state from file to file?! Because that would suck!

I bet this is what was actually causing our self referencing rules.

Just as one more annoying check, I wonder if this would fix all of our problems on it's own (without the global change). Would you mind trying? 

It just makes me feel a bit odd that it works in Storybook but not from another project.

fallaciousreasoning (commented):
This mostly looks pretty good to me just one thing I think we should try before merging it:

I have a suspicion that just your last bit (the `rules = {}` `nodesToDelete.clear()`) could fix our layering problem

postcss/theme.js
fallaciousreasoning: NVM, I just saw the `rules = {}` at the bottom

postcss/theme.js
AlanBreck: Removed this logic in favor of leaning on the reset of `rules`.

postcss/theme.js
AlanBreck: While not solving the same issue as solved by `useGlobal`, this does negate the need for `hasThemeAtRule`, so that's been removed.

fallaciousreasoning (approved):
LGTM, thanks for doing this @AlanBreck

PR #44, opened 2022-10-06
Add script for generating React bindings

fallaciousreasoning:
This is based on the `genprops` script. In addition, this PR removes the generated files from the repository and regenerates them as part of the `build` script.

PR #79, opened 2022-11-11
WIP: adds input components

AlanBreck:
(partially) Resolves https://github.com/brave-experiments/leo/issues/54

I started this PR without realizing we had an existing one here: 
https://github.com/brave-experiments/leo/pull/30

I think we may want to go with this one since it leverages default input props, but I also copied 
some things from the other PR which were more elegant. There may still be more that needs to be 
replicated here, or we can scratch entirely.

This is notably missing icons, which is somewhat intentional as I believe we need to sort that 
out first.

### DIFF
--- a/web-components/input/text/react.ts
+++ b/web-components/input/text/react.ts
@@ -1,0 +1,9 @@
+import SvelteToReact, { ReactProps } from "../../svelte-react";
+import TextInput from './text.svelte'
+import type { TextInputEvents, TextInputProps } from './text.svelte'
+
+// Prop types for a svelte component are not able to be inferred at compile-time,
+// only at dev time in typescript plugins, so we must duplicate it here, manually.
+// TODO: have this type definition be auto-generated in the bundler.
+
+export default SvelteToReact<ReactProps<TextInputProps, TextInputEvents>>('leo-text-input', TextInput)
--- a/web-components/input/text/text.svelte
+++ b/web-components/input/text/text.svelte
@@ -1,0 +1,117 @@
+<svelte:options tag="leo-text-input" />
+
+<script lang="ts">
+  import type * as Props from './props';
+
+  interface CommonProps {
+    type?: Props.TextInputType;
+  }
+
+  type TextInputProps = CommonProps & Partial<svelte.JSX.HTMLAttributes<HTMLElementTagNameMap['input']>> & {
+    isRequired?: boolean;
+    isDisabled?: boolean;
+  }
+
+  type $$Props = TextInputProps
+
+  export let value: string = '';
+  export let name: string = '';
+  export let label: string = '';
+  export let error: string = '';
+  export let type: Props.TextInputType = 'text';
+  export let isRequired: boolean = false;
+  export let isDisabled: boolean = false;
+</script>
+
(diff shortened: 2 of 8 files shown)


web-components/input/text/react.ts
```diff
@@ -0,0 +1,9 @@
+import SvelteToReact, { ReactProps } from "../../svelte-react";
+import TextInput from './text.svelte'
+import type { TextInputEvents, TextInputProps } from './text.svelte'
+
+// Prop types for a svelte component are not able to be inferred at compile-time,
+// only at dev time in typescript plugins, so we must duplicate it here, manually.
+// TODO: have this type definition be auto-generated in the bundler.
```
fallaciousreasoning, line 7: We actually have a PR that does this: https://github.com/brave-experiments/leo/pull/44

I think ideally we should land #44 first and then use it's auto generated React binding.

web-components/input/text/text.svelte
```diff
@@ -0,0 +1,117 @@
...
+  interface CommonProps {
+    type?: Props.TextInputType;
+  }
+
+  type TextInputProps = CommonProps & Partial<svelte.JSX.HTMLAttributes<HTMLElementTagNameMap['input']>> & {
+    isRequired?: boolean;
+    isDisabled?: boolean;
+  }
+
+  type $$Props = TextInputProps
```
fallaciousreasoning, line 15: I think this might be my least favorite thing about Svelte 😢

fallaciousreasoning (commented):
Looks like a pretty sensible approach to me!

web-components/input/text/text.svelte
```diff
@@ -0,0 +1,117 @@
...
+
+    &:disabled {
+      --text-color: var(--color-text-disabled);
+      --placeholder-text-color: var(--color-text-disabled);
+      --bg: var(--color-container-disabled);
+    }
+  }
+
+  // TODO: is this the best approach?
+  :global(.form-control) + :global(.form-control) {
```
petemill, line 114: Perhaps we need one of the following:
- an independent `.form-item` class that can be imported by this component and any others (including "third party" custom items)
- a separate `form-item.svelte` component that we wrap this and other form elements with 
- Let the consumer deal with spacing by adding their own class names or wrapping flex / grid parents
- Leo provides a FormItemGroup component that provides a default flex parent with defined gaps.

I would also suggest that this should be margin and not padding, unless I'm otherwise mistaken.

web-components/input/text/text.svelte
```diff
@@ -0,0 +1,117 @@
...
+  export let label: string = '';
+  export let error: string = '';
+  export let type: Props.TextInputType = 'text';
+  export let isRequired: boolean = false;
+  export let isDisabled: boolean = false;
+</script>
+
+<div class="form-control" class:required={isRequired} class:error={error} class:disabled={isDisabled}>
+  {#if label}
+    <label for={name}>{label}</label>
```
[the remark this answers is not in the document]
AlanBreck: I was assuming use in a standard form, which would require a `name` attribute, but I suppose these could be used outside of a `form` quite a bit.

AlanBreck:
> an independent .form-item class that can be imported by this component and any others (including "third party" custom items)

@petemill, are you referring to CSS-modules here?

PR #100, opened 2022-11-22
[Component] Leo Icon

### LINKED ISSUE #45: Icons
### Serving / Requesting
We want to allow:
- Consumers to self-host all the .svg files so that they can reduce duplication among bundles (e.g. brave-core serving icons from chrome://resources/leo-icons/x.svg
  - Consumers could choose to only serve .svg files they are using, to conserve package size
  - Consumers could serve icons from a central URL but Leo JS from local bundles
- Consumers can define a custom library of .svg files, and get the benefits of a `LeoIconComponent`

All icons exist in repo as separate .svg files, e.g. `/leo/icons/xyz.svg`
`LeoIconComponent` fetches the file and renders the svg, adds css class to it.
Consumer passes in icon-name and 
```
<LeoIcon library="leo" label="" name="search" />
````

Take inspiration from https://shoelace.style/components/icon / https://github.com/shoelace-style/shoelace/tree/next/src/components/icon

### Benefits of `LeoIconComponent`
- Change fill / stroke to `currentColor` so it inherits `color` via css
- Restrict width and height to `--leo-icon-size` so that the icon fits in to containers, e.g. different button types
- Fetches and inlines

### What it won't do
- Accept custom color property, this should be cascaded through css, i.e. set `color: pink` on the custom element or container or anywhere. Same with `--leo-icon-size`.

### PR DESCRIPTION
**Pull request by fallaciousreasoning**
Fixes #45 

Storybook: https://leo-storybook-git-leo-icon-fallaciousreasoning.vercel.app/?path=/story/icon--default

This adds a `leo-icon` component. Currently the following is supported:

- Set the color via `color: red`
- Don't override color of `-color` icons (this relies on the naming convention remaining the same)
- Set size via `--leo-icon-size`

In addition, this package adds support for downloading all the icons from Figma via the `npm run update-icons` command, using a fork of https://github.com/slawomirkolodziej/figma-api-exporter

~~However, the licensing for that package is unclear. I've requested clarification (https://github.com/slawomirkolodziej/figma-api-exporter/issues/5) and if they don't add a license we might have to roll our own, which would be a bit sad but not too difficult.~~ ~~The author is pretty responsive, and I think it's going to be MIT licensed!~~ The package is MIT licensed!

Note: We're pointing at a revision on the Brave fork of the library for security: https://github.com/brave/security/issues/1137#issuecomment-1341769048 

[Screencast from 2022-11-23 12-16-53.webm](https://user-images.githubusercontent.com/7678024/203440441-2e5da95e-0a4e-4f7e-ac02-2bb58af004c5.webm)

### CODE REVIEW

**Comment by petemill** [comment]
@AlanBreck and I were talking about the colors and whilst the mutate function is great to have, perhaps it would be more efficient (as well as allow some non-js implementations) if we stripped the fill attributes in the update-icons.js script?

**Comment by fallaciousreasoning** [comment]
Also flags don't have the color suffix so we should special case them (and possibly rename them so we can render the icons with `Country=NZ` instead of `Country=NZ New Zealand (Aotearoa)`

**Comment by fallaciousreasoning** [comment]
Okay, I've pushed the following changes:
- Mutate the icons when we download them instead of at runtime.
- Country Icons are always treated as color icons
- Added support for icons via a slot, as well as via the name attribute, which solves @petemill's `import` problem.
- Added a story for custom icons

File: package.json [hunk 1]
```diff
@@ -33,6 +34,7 @@
     "@types/jest": "29.2.4",
     "babel-loader": "9.1.0",
     "css-loader": "6.7.2",
+    "figma-api-exporter": "github:slawomirkolodziej/figma-api-exporter#664caaa010f295bde15ec4db752abd0a2cc8d2af",
```
AlanBreck, line 37: Do we need a sec review on this one?

web-components/icon/Icon.svelte
```diff
@@ -0,0 +1,74 @@
+<svelte:options tag="leo-icon" />
+
+<script context="module" lang="ts">
```
AlanBreck, line 3: Curiosity more than anything: why is this code in `context="module"` and not the main `script` tag?

web-components/icon/Icon.svelte
```diff
@@ -0,0 +1,74 @@
...
+    const url = library.resolve(name)
+    const response = await fetch(url)
+    if (!response.ok) return { ok: false, svg: null }
+
+    const parsed = parser.parseFromString(await response.text(), 'text/html')
+    const svgRoot = parsed.querySelector('svg')
+    if (!svgRoot) return { ok: true, svg: '' }
+
+    // If the library has a mutate function, run it.
+    library.mutate?.(svgRoot, name)
```
AlanBreck, line 21: I haven't seen this syntax before! How does this work?

web-components/icon/library.default.ts
```diff
@@ -0,0 +1,11 @@
+import { IconLibrary, registerLibrary } from "./iconLibrary";
+
+let defaultIconsPathBase = './icons';
+export const setDefaultIconsPath = (iconsPath: string) => defaultIconsPathBase = iconsPath;
```
AlanBreck, line 4: How is this meant to be used?

AlanBreck (commented):
Sorry for some of the curiosity questions. Two more questions, though:
1. Can we make the name case insensitive? E.g. `ATM` and `atm` both work?
2. Setting `library` knob in Storybook throws an error.

package.json
fallaciousreasoning: We do indeed - it's over here https://github.com/brave/security/issues/1137
Includes the dependency & the `@html` review :smile:

web-components/icon/Icon.svelte
fallaciousreasoning: It's null propagation - essentially only call the method if it's not null

`ts
const f = foo?.bar?.baz

// is the same as
const f = foo
	? foo.bar
		? foo.bar.baz
        : undefined
    : undefined

// you can also use it for arrays/funcs but the syntax is a bit more funky

const nthItem = items?.[n]
// which is the same as 
const nthItem = items ? items[n] : undefined

// and for functions, they'll only be called if they're defined
const result = func?.(stuff, things)

// which is
const result = func ? func(stuff, things) : undefined

web-components/icon/library.default.ts
fallaciousreasoning: Basically, you can use it to set where the icons are served from. I.e. if we wanted to serve them from `chrome://resources/leo-icons` we could call `setDefaultIconsPath('chrome://resources/leo-icons')` and it would fetch the `atm` icon from `chrome://resources/leo-icons/atm.svg`

web-components/icon/Icon.svelte
fallaciousreasoning: So that's so the cache is shared between instances of the icon component. If it was in the main script tag each icon would fetch the SVG again.

fallaciousreasoning:
> Setting library knob in Storybook throws an error.

FWIW, that's expected - only the default library is defined and trying to use a non-existent library probably should break, as it wouldn't display anything, so is probably a mistake.

> Can we make the name case insensitive? E.g. ATM and atm both work?

We definitely can - we could lowercase all icon names. I was thinking of creating a typescript file with all the leo icons in it though, so we get intellisense. No strong feelings on caseSeNsiTivitY though. @petemill do you have a preference?

web-components/icon/Icon.svelte
AlanBreck: Super cool.

AlanBreck (approved):

fallaciousreasoning:
Currently, we have a few actions to make for the Security Review:
- Add DOMPurify to the icon component.
- Depend on a fork of https://github.com/slawomirkolodziej/figma-api-exporter to guard against potentially malicious changes.

web-components/icon/icon.svelte
```diff
@@ -0,0 +1,56 @@
...
+
+    & .icon, svg {
+      width: 100%;
+      height: 100%;
+    }
+
+    /* Non color icons are set via a mask-image, so we can change the color
+     * by setting the background */
+    & .icon:not(.color) {
+      background: var(--icon-color);
```
petemill, line 43: How about we use `background-color: currentColor` and then we can just rely on color cascading for both inline and background variants?

fallaciousreasoning: I had no idea you could do this :O I thought `currentColor` was a special SVG thing ! Awesome!

thypon (approved):

AlanBreck (approved):

package.json
```diff
@@ -6,11 +6,14 @@
   "dependencies": {
     "@ctrl/tinycolor": "3.5.0",
     "@tsconfig/svelte": "3.0.0",
+    "@types/dompurify": "^2.4.0",
+    "dompurify": "^2.4.1",
+    "jsdom": "^20.0.3",
```
petemill, line 11: Everything else in dependencies has the version pinned, so seems like we should continue to do that for these new ones

petemill (approved):

fallaciousreasoning:
I'm holding off on the merge till the bundling PR #161 lands, as it has priority, and this is huge. (sorry for all the force pushes, I accidentally push the bundling PR to the icons branch :fearful: )

PR #101, opened 2022-11-23
[Meta] Add prettier for formatting.

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/102
This adds prettier, with the same config we're expecting to have on Brave Core. In addition, the format is run as part of CI, and will fail if code is not correctly formatted.


.github/workflows/build.yml
```diff
@@ -30,4 +30,9 @@ jobs:
...
         id: build-leo
         run: |
-          npm run transform-tokens
\ No newline at end of file
+          npm run transform-tokens
+
+      - name: Format
+        id: format
+        run: |
+          npx prettier ./ --list-different
```
petemill, line 38: Since we're specifying `prettier` as a dependency, can we run it via `node` / `npm` instead of `npx`?

examples/example-ui-react/src/MySvelteComponent/react.ts
```diff
@@ -1,13 +1,14 @@
...
 import Button from './button.svelte'
 
 // Prop types for a svelte component are not able to be inferred at compile-time,
 // only at dev time in typescript plugins, so we must duplicate it here, manually.
 // TODO: have this type definition be auto-generated in the bundler.
 
-export interface HelloEvent extends CustomEvent<{
-  text: string
-}> {}
+export interface HelloEvent
```
petemill, line 8: IMO this example is less readable, but that's minor!

petemill (approved):
Thanks!!

examples/example-ui-react/src/MySvelteComponent/react.ts
fallaciousreasoning: Yup, I tend to agree.

.github/workflows/build.yml
fallaciousreasoning: We can yeah. I've added a `format-diff` script which we can run during the build workflow. FWIW, I'm pretty sure `npx` will use the installed version of the package, and if it isn't installed, it will ask if you want to download it.

.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "trailingComma": "none",
+  "tabWidth": 2,
+  "semi": false,
```
AlanBreck, line 4: No semicolons??

...I'll bend to the will of the people.

fallaciousreasoning: It makes me sad too. Just wanted it to be consistent with the Brave Core one.

@petemill?

petemill: We already have no semicolons in this repo and most(?) of Brave JS repos - blame StandardJS!

fallaciousreasoning:
@petemill I've rebased this and updated everything. Hopefully we can land this!

AlanBreck (approved):

Issue #65, opened 2022-11-02
Component: Navigation dots


aguscruiz:
https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?node-id=200%3A11392

![image](https://user-images.githubusercontent.com/15734630/199499953-abe03c31-cfc6-480f-bbf6-30f8592fa11a.png)

fallaciousreasoning:
It'd be cool to use this on the New Tab Page for the Grid of sites (there's a similar control there but it would be great to be able to use Leo

fallaciousreasoning:
Hey @aguscruiz I've just started implementing this and I have a couple of questions:
1. Can we make the hover state more obvious? Currently it's nearly identical to the Default state.
2. Should the pill (for the current dot) have a hover state or a `focus:visible` state? Interacting with it is a no-op.

aguscruiz:
Looks great! Answering your questions:
1- Already fixed on Figma, let me know what you think
2- Also answered, I think it shouldn't have a hover state, but it should probably have a focus one.
3- That's what I was thinking, I would add an easing property to make it a bit smoother. `ease-in-out` should improve it a lot, let's try it out. And in general, ease-in-out should be the default easing I think (I'm working on drafting that spec)

PR #107, opened 2022-11-27
[Component] Navdots

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/65

Add navigation dots
[Screencast from 2022-11-28 12-34-57.webm](https://user-images.githubusercontent.com/7678024/204165735-52252112-89e6-4a8d-8176-6ed5e6d6f785.webm)

Storybook:
https://leo-storybook-git-wc-nav-dots-fallaciousreasoning.vercel.app/?path=/story/nav-dots--default

### DIFF
--- a/web-components/navdots/navdots.stories.svelte
+++ b/web-components/navdots/navdots.stories.svelte
@@ -1,0 +1,23 @@
+<script lang="ts">
+  import { Meta, Story, Template } from '@storybook/addon-svelte-csf'
+  import NavDots from './navdots.svelte'
+
+  let activeDot = 0
+  function handleChange(n) {
+    activeDot = n.detail.activeDot
+  }
+</script>
+
+<Meta
+  title="Nav Dots"
+  component={NavDots}
+  argTypes={{
+    dots: { control: 'number' },
+  }}
+/>
+
+<Template let:args>
+  <NavDots dotCount={args.dots} activeDot={activeDot} on:change={handleChange}/>
+</Template>
+
+<Story name="Default" />
--- a/web-components/navdots/navdots.stories.svelte
+++ b/web-components/navdots/navdots.stories.svelte
@@ -13,12 +13,16 @@
   component={NavDots}
   argTypes={{
     dots: { control: 'number', defaultValue: 10 },
-    dotSize: { control: 'number', defaultValue: 8 }
+    dotSize: { control: 'number', defaultValue: 8 },
+    activeColor: { control: 'color' },
+    defaultColor: { control: 'color' },
   }}
 />
 
 <Template let:args>
-  <NavDots dotCount={args.dots} activeDot={activeDot} on:change={handleChange}/>
+  <div style={`--color-interaction-button-primary-background: ${args.activeColor || 'unset'}; --color-primary-20: ${args.defaultColor || 'unset'}`}>
+    <NavDots dotCount={args.dots} {activeDot} on:change={handleChange} />
+  </div>
 </Template>
 
 <Story name="Default" />
--- a/web-components/navdots/navdots.stories.svelte
+++ b/web-components/navdots/navdots.stories.svelte
@@ -12,7 +12,8 @@
   title="Nav Dots"
   component={NavDots}
   argTypes={{
-    dots: { control: 'number' },
+    dots: { control: 'number', defaultValue: 10 },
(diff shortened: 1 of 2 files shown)


aguscruiz:
@fallaciousreasoning should I wait for the latest updates to review? Let me know

aguscruiz:
Minor thing that's not gonna be in the design, the cursor should be the pointer for these buttons (and in general, I think a good rule is making all the clickeable elements have the pointer cursor)

fallaciousreasoning:
Nah, I think any feedback you have will be useful now (the cursor is a good catch, I completely forgot :D)

aguscruiz:
Ok, I'll add it here then: 

- The inactive dot should be 8px wide/tall. And has a 1px margin, so it's 10px wide/tall counting it.
- Active dot is 10px tall, a bit bigger than the inactive, and 18px wide.
- Margin between dots (without counting the active dot) is 8px.
- Adjusted the hover states in the design based on your comments.
- Have the cursor be pointer
- I think the animation works perfect here 

image
- I don't know if this nav margin is a default or just for storybook, but I'd suggest not adding it by default because it may change depending on the design.

fallaciousreasoning:
>- The inactive dot should be 8px wide/tall. And has a 1px margin, so it's 10px wide/tall counting it.

Whoops - I made it double size to make it easier to test and forgot to change it back!

> - Active dot is 10px tall, a bit bigger than the inactive, and 18px wide.

I completely missed the fact it was taller than the dots! Fixed.

>- Adjusted the hover states in the design based on your comments.

Cheers, thank you!

>- Have the cursor be pointer

Nice catch!

>- I think the animation works perfect here 

:heart: 

>- I don't know if this nav margin is a default or just for storybook, but I'd suggest not adding it by default because it may change depending on the design.

Oooh, didn't catch that. I swapped to using an `<ol>` tag for accessibility and I think the margin came with it. Removed!

aguscruiz (approved):

web-components/navdots/navdots.stories.svelte
```diff
@@ -0,0 +1,28 @@
...
+  argTypes={{
+    dots: { control: 'number', defaultValue: 10 },
+    dotSize: { control: 'number', defaultValue: 8 },
+    activeColor: { control: 'color' },
+    defaultColor: { control: 'color' },
+  }}
+/>
+
+<Template let:args>
+  <div style={`--color-interaction-button-primary-background: ${args.activeColor || 'unset'}; --color-primary-20: ${args.defaultColor || 'unset'}`}>
```
petemill, line 23: What do you think about making a component-specific variable that can be overriden, e.g. `--color-navdots`. I worry about the parents knowing the specific variable that gets used.

petemill: Separately, and this is also a question for @aguscruiz - do we need to allow overriding _active_ dot color as well as dot color, or is there some kind of consistent calculation from dot-color to active-dot-color, like opacity?

fallaciousreasoning:
- Move CSS variables to root & make them global
- Maybe use a span for the active indicator

web-components/navdots/navdots.svelte
```diff
@@ -0,0 +1,127 @@
...
+      background: var(--color-primary-20);
+      transition: background-color var(--transition-duration)
+          var(--transition-easing),
+        box-shadow var(--transition-duration) var(--transition-easing);
+
+      &:hover {
+        background-color: var(--color-primary-30);
+      }
+
+      &:focus-visible:not(.active) {
```
petemill, line 99: let's wrap this with 
```
@supports selector(:has(a)) {
}
```

web-components/navdots/navdots.svelte
```diff
@@ -0,0 +1,127 @@
...
+      position: relative;
+      padding: 0 calc(var(--dot-spacing) / 2);
+      margin: 0;
+      list-style: none;
+
+      > li {
+        display: flex;
+      }
+
+      &:has(.dot.active:focus-visible) .active-dot {
```
petemill, line 77: let's wrap this with 
```
@supports selector(:has(a)) {
}
```

web-components/navdots/navdots.stories.svelte
fallaciousreasoning: Yeah, it'd be really neat if everything derived from one color :smile:

PR #112, opened 2022-12-01
moves svelte2tsx to prod dep

AlanBreck:
Installing in another project was returning an error due to a this package missing.


fallaciousreasoning (approved):
I guess this is happening because we run the `gen-types` & `gen-react` scripts at install time. LGTM

PR #135, opened 2022-12-14
Separate Button into a separate Link and Button component to fix typings.

fallaciousreasoning:


AlanBreck:
No!! :-(

Not happy with Svelte right now.

fallaciousreasoning:
I can get the typings working for the `anchor`/`button` component but then it breaks the React wrapper :D I can have a bit more of a dig if you like?

fallaciousreasoning:
Have a look at #136 and see if that looks better? Both ways aren't ideal, but hopefully the solution in #136 disappears into the background a bit more

fallaciousreasoning:
Closing in favor of #136

PR #136, opened 2022-12-15
Fix React bindings when we have generic arguments

fallaciousreasoning:
This is an alternative to #135, which gets generic type definitions working nicely with React. Made in response to https://github.com/brave/leo/pull/135#issuecomment-1352355559

### DIFF
--- a/web-components/button/button.svelte
+++ b/web-components/button/button.svelte
@@ -47,7 +47,7 @@
   export let isDisabled: Disabled = undefined
   export let href: Href = undefined
 
-  const tag = href ? 'a' : 'button'
+  $: tag = href ? 'a' : 'button'
 
   const dispatch = createEventDispatcher()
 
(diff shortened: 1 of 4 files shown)


web-components/button/button.svelte
```diff
@@ -17,51 +18,73 @@
...
+</svelte:element>
+
 <style lang="scss">
   // Main styles and states
   .leoButton {
     // Gradients cannot have a transition, so we need to reset `transition`
     // to only apply to `box-shadow` and `border-color` in .isCTA
-    --default-transition: box-shadow 0.12s ease-in-out, color 0.12s ease-in-out, border-color 0.12s ease-in-out;
+    --default-transition: box-shadow 0.12s ease-in-out, color 0.12s ease-in-out,
+      border-color 0.12s ease-in-out;
```
AlanBreck, line 87: Ooof... not cool, Prettier. Not cool.

AlanBreck (approved):
Looks good to me, but I'm also biased towards keeping the anchor and button in one component. :-P

fallaciousreasoning:
All right, that's good enough for me. I'm going to leave it as is, and we can revisit if it becomes a problem.

Closes #135

PR #151, opened 2023-01-06
Fix `svelte-check` errors and add check to CI

fallaciousreasoning:
I merged the `svelte-check` PR here but the new `svelte-check` is a bit stricter and I noticed it was complaining about some previously valid code.

In addition, I'm running `svelte-check` as part of CI, so we'll catch regressions. In future, it'd be good to enable the `--fail-on-warnings` flag but at the moment there's no way of whitelisting our `@theme` at rule

### DIFF
--- a/web-components/navdots/navdots.svelte
+++ b/web-components/navdots/navdots.svelte
@@ -23,13 +23,12 @@
 </script>
 
 <nav class="leo-navdots" aria-label={label}>
-  <ol class="dot-container" part="dot-container" bind:this={container}>
+  <ol class="dot-container" bind:this={container}>
     {#each dots as dot}
       <li>
         <button
           class="dot"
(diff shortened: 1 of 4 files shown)


fallaciousreasoning:
@AlanBreck could you PTAL?

web-components/navdots/navdots.svelte
```diff
@@ -23,21 +23,20 @@
 </script>
 
 <nav class="leo-navdots" aria-label={label}>
-  <ol class="dot-container" part="dot-container" bind:this={container}>
```
fallaciousreasoning, line 26: Note: I had a talk with Pete a while back and we're not going to expose `part`s until we need them :smile: In addition, these were failing the new version of Svelte Check

AlanBreck (approved):

Issue #16, opened 2022-08-10
What does compiled output look like for web projects


petemill:
If we wanted to consume leo components from other projects with minimal bundler / build configuration, what would that look like? What are some svelte examples to follow?

We would want to keep:
- Type safety and intellisense for react consumers (e.g. button/react.ts)
- Type safety and intellisense for svelte consumers (e.g. button/button.svelte)

fallaciousreasoning:
I think #26 should solve this. It uses one of the internal Svelte projects ([svelte2tsx](https://github.com/sveltejs/language-tools/tree/master/packages/svelte2tsx)) to generate type definitions as part of the build process.

fallaciousreasoning:
A problem we have with this currently is that we compile Leo in consuming projects, which means that projects which use Leo need to depend on Svelte and PostCSS.

That said, maybe this is okay? We could solve by precompiling before publishing to NPM but there are definitely tradeoffs.

AlanBreck:
What are the downsides to precompiling?

fallaciousreasoning:
I guess we could get multiple copies of things? It's probably fine with Svelte though as the library gets compiled away. I guess the main one would be we need to remember to compile it 😆

petemill:
One downside could be more difficult local development workflow if the directory structure differs between source and packaged npm

fallaciousreasoning:
Presumably, if it compiled to a `./build` folder or something, we could point to that folder. It might make pointing to a Github repository harder though.

PR #161, opened 2023-01-17
Add support for bundling Leo using Rollup

fallaciousreasoning:
This will let other packages consume Leo without needing to depend on Svelte or PostCSS. As a proof of concept I've removed the Svelte deps from the `example-ui-react` project and added a new `plain-html` example which uses the raw web-components.

In addition I've updated the paths to make imports nicer. The new folder layout is something like this:
```
leo/ # Root folder
    web-components/ # All compiled web-components, Generated
    react/ # All react bindings
    svelte/ # preprocessed svelte files, for use in svelte projects
    tokens/css/ # Compiled css
    tokens/ios/ # Compiled ios
    tokens/ios-swift/ #compiled ios-swift
    tokens/json/ # ./build/json
    tokens/skia/ #./build/skia
    tokens/tailwind/ # ./build/tailwind
    icons/ # All icons
    src/ # Everything we need to generate stuff
        components/ # Where we write our Svelte components
        tokens/ # The current ./tokens folder
        postcss/ # Our postcss plugins
        scripts/ # Our build scripts
    tests/ # Not moved
    examples/ # Not moved
```

Once this PR is finished & discussed, it should fix https://github.com/brave/leo/issues/16

Note: This PR doesn't preprocess the Svelte files and drop them into the `svelte` folder. That can be done separately to this PR, as the only current Svelte consumers are preprocessing the source files from `./src`.


petemill:
This is actually great - I was originally thinking this complicates things as we have many different types of consumers and workflows. But I think you've shown that we can get something that would have few disadvantages. Seems like there couple be these things to solve:

## Paths for import
Perhaps the clearest expected paths for import from various languages would be something like:
  - `@brave/leo/css/variables.css`
  - `@brave/leo/skia/colors.h`
  - `@brave/leo/web-components/icon`
If we wanted to accomplish this then we could move all source code to `/src`? We could also publish mainly only the build directory, but then that causes an issue with a static directory like /icons. We'd also have to copy the package.json and README.md files (and some others) and that can get messy, unless the tooling is better these days.

## Ensure consuming bundlers don't double-up dependencies
If this repo is outputting a JS bundle for each component, we want to make sure that we still benefit from:
- Sharing imported instances of node_module dependencies, instead of bundling multiple copies separately (e.g. react, svelte).
- Sharing Leo-specific code between Leo components on the same page.
We can assume that the consumer will be using some kind of bundler that uses es module import syntax. We could also assume that if no bundler is being used then the platform supports es module import syntax anyway. So perhaps we can leave as many es-module imports as possible there. And on that note it seems this is generating two styles - es modules at .mjs and amd modules at .js. Perhaps we keep it simple and only generate es-modules?
In this PR that seems mostly accomplished - in the output I see `import xyz from 'react' as well as shared imports between the components from '../index-xyz.js`. That file though seems to have the svelte runtime stuff. I wonder if that can be imported from somewhere that a consuming app could use, or maybe svelte isn't compatible with that kind of thing as all the functions are generated?

## Developer workflow
I imagine I'd build some Leo changes in the context of a consuming app. Presumably I could still do this via something like `npm link` if we
- Did not publish a different directory structure than is generated in the context of the repo
- Had some kind of file watch on the source code that rebuilds via rollup. I assume rollup already has some kind of watch functionality like this anyway. We may be concerned with the time it takes to generate the entire bundle when just modifying 1 component, but we're not at that scale yet and rollup probably thought about that. The only issue thing we may want to validate is if this works in the context of a consuming app which would also be watching the output files. I think the answer is that it would work fine because the imported file names would be static and not the generated index-1234.js names.

petemill:
One more question I have about this compiled output is whether it's directly consumable by a Svelte app as Svelte components?

package.json
```diff
@@ -47,7 +51,7 @@
     "svelte-loader": "3.1.4"
   },
   "peerDependencies": {
-    "typescript": "4.9.4"
+    "typescript": "^4.7.0"
```
AlanBreck, line 54: Why are we reverting to a previous version here?

fallaciousreasoning: It's a peer dependency, and we want to be as permissive as possible here

src/package.json
```diff
@@ -0,0 +1 @@
+{ "type": "commonjs" }
```
fallaciousreasoning, line 1: Note: This let's us write commonjs in `src/` which is what all our transformations and scripts are using, while exporting ES Modules.

src/scripts/gen-react-bindings.js
```diff
@@ -47,25 +47,30 @@ const getReactFileContents = async (svelteFilePath) => {
         ? `<${generics.map(g => g.name).join(', ')}>`
         : ''
 
-    const fileContents = `
-import type * as React from 'react';
-import SvelteToReact, { type ReactProps } from '${svelteReactWrapperRelativePath}';
-import ${componentName} from './${fileName}';
-import type { ${componentName}Events as SvelteEvents, ${componentName}Props as SvelteProps } from './${fileName}';
+    const binding = `
```
fallaciousreasoning, line 50: I split this into a separate declaration & bindings file because it makes the build process simpler, and it means we don't have to do the monster `as unknown as CrazyType` we were doing before.

src/scripts/gen-react-bindings.js
```diff
@@ -75,4 +80,9 @@ const createReactBindings = async (rootDir) => {
     }
 }
 
-createReactBindings('./web-components')
+module.exports = createReactBindings
+
+if (require.main == module) {
```
fallaciousreasoning, line 85: These run automatically as part of rollup now. As a side effect it means you can run
`npx rollup -c rollup.config.js --watch` and it will automatically watch for changes!

AlanBreck (approved):

examples/plain-html/README.md
```diff
@@ -0,0 +1,8 @@
+# An HTML + JS example application using Leo components
+
+To run
+
+    npx serve
+
+Note: the `build` folder here is a symlink to the build folder from Leo, so we
```
petemill, line 7: nit: no longer the `build` folder

petemill (approved):
Nice rollup config!

package.json
```diff
@@ -34,25 +37,35 @@
...
+      "optional": true
+    }
   },
   "scripts": {
-    "install": "npm run build",
-    "start": "npm run build",
-    "build": "npm run transform-tokens && npm run gen-types && npm run gen-bindings",
-    "gen-types": "node ./scripts/gen-svelte-types.js",
-    "gen-bindings": "node ./scripts/gen-react-bindings.js",
+    "prepare": "npm run build",
```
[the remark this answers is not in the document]
fallaciousreasoning: Okay, so I tested with `install` - it doesn't have `devDependencies` available, unlike `prepare`.

Issue #63, opened 2022-11-02
Component: Label


aguscruiz:
https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?node-id=195%3A15172

![image](https://user-images.githubusercontent.com/15734630/199499597-151d1534-2ca7-4f0f-8ce7-3282c17ec861.png)

OAuthBringer:
Required for Creators 2.0

PR #180, opened 2023-01-26
[Component]: Label

fallaciousreasoning:
Adds a Label component
Fixes https://github.com/brave/leo/issues/63

~~**Note:** The hero color doesn't work because it requires #75~~

### DIFF
--- a/src/components/label/label.svelte
+++ b/src/components/label/label.svelte
@@ -1,0 +1,32 @@
+<script context="module" lang="ts">
+    const colors = {
+        
+    }
+</script>
+<script lang="ts">
+    export let mode: 'faint' | 'heavy' = 'faint'
+    export let color: 'gray' | 'secondary' | 'primary' | 'red' | 'yellow' | 'green' | 'blue' | 'purple' | 'pink' | 'teal' | 'hero' = 'gray'
+
+    $: background = color === 'hero'
+        ? `var(--gradient-hero)`
+        : mode === 'heavy'
+            ? `var(--color-${color}-40)`
+            : `var(--color-${color}-10)`
+    $: text = color === 'hero'
+        ? `var(--color-white)`
+        : mode === 'heavy'
+            ? `var(--color-${color}-10)`
+            : `var(--color-${color}-50)`
+</script>
+
+<style lang="scss">
+    .leo-label {
+        display: inline;
+        border-radius: var(--radius-4);
+        padding: var(--spacing-4);
+    }
+</style>
+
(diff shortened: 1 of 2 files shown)


fallaciousreasoning:
So I was just having a look at our variables - it looks like the font-weight for `--font-text-x-small-regular` is `400` and we don't have any separate variables for `font-weight`. Do you think we should add some?

Also, I don't know if we can (or should) have separate styles for Apple devices. I had a quick look and it doesn't seem like there's any easy way of doing it. @petemill do you have any suggestions?

AlanBreck:
I would agree that it's best if we can avoid having device specific styling. One thing we [_could_](https://www.youtube.com/watch?v=g3j9muCo4o0) do is to add CSS rules for things like `.is-macos` or `.is-ios`, and then leave the actual OS detection up to the client, and inform them that they can add these classes (like `.is-macos`) to the body.

src/components/label/label.svelte
```diff
@@ -0,0 +1,67 @@
...
+<div class="leo-label" style:background style:color={text}>
+  <div class="content">
+    <slot>Label</slot>
+  </div>
+</div>
+
+<style lang="scss">
+  .leo-label {
+    --icon-size: var(--leo-label-icon-size, 13px);
+    --font-text: var(--leo-label-font-text, var(--font-text-x-small-semibold));
```
petemill, line 48: The figma now uses `--font-components-label`

src/components/label/label.svelte
```diff
@@ -0,0 +1,67 @@
...
+    'secondary',
+    'primary',
+    'red',
+    'yellow',
+    'green',
+    'blue',
+    'purple',
+    'pink',
+    'teal',
+    'hero'
```
petemill, line 13: `hero` as a color should actually be `CTA`. `heavy` mode should be called `hero`

src/components/label/label.svelte
```diff
@@ -0,0 +1,67 @@
...
+  ] as const
+  export type Color = typeof colors[number]
+
+  export const modes = ['faint', 'heavy'] as const
```
petemill, line 17: `modes` should be renamed to `'loud' | 'default'` cc @aguscruiz

petemill (approved):

src/components/label/label.svelte
fallaciousreasoning: Renamed to `CTA`

src/components/label/label.svelte
fallaciousreasoning: Renamed - defaults to `default`

PR #184, opened 2023-01-26
Ignore install scripts for npm packages

AlanBreck:
This allows us to avoid running install scripts, like [this one from `core-js-pure`](https://github.com/brave/leo/pull/181#issuecomment-1405099072).

This protects us [against some supply chain attacks](https://dev.to/naugtur/get-safe-and-remain-productive-with-can-i-ignore-scripts-2ddc). We can periodically use [can-i-ignore-scripts](https://www.npmjs.com/package/can-i-ignore-scripts) to determine if we need to allow any more by including in the `npm build ...` line.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,9 @@ jobs:
             ${{ runner.os }}-npm-
 
       - name: Install dependencies
-        run: npm ci
+        run: |
+          npm ci --ignore-scripts
+          npm rebuild style-dictionary-create-react-app
 
       - name: Build Leo
         id: build-leo
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,6 +42,7 @@ jobs:
       - name: Install dependencies
         run: |
           npm ci --ignore-scripts
+          # Please check with Brave's security team prior to adding any packages to this list.
           npm rebuild style-dictionary-create-react-app
 
       - name: Build Leo


thypon (approved):

petemill (approved):
This repo itself uses a necessary install script so perhaps we need to add it 😄 . Though the problem with [can-i-ignore-scripts](https://www.npmjs.com/package/can-i-ignore-scripts) seems to be that it's not that extensive? And once a repo is approved then it could still be hijacked before that repo is updated? Our pinned dependency versioning should help with that.

AlanBreck:
> This repo itself uses a necessary install script so perhaps we need to add it 😄 .

This repo wouldn't have to add itself, right? Consuming repos would have to allow leo to run scripts, if otherwise disabled with `--ignore-scripts`, right?

> Though the problem with [can-i-ignore-scripts](https://www.npmjs.com/package/can-i-ignore-scripts) seems to be that it's not that extensive?
In what way?

> And once a repo is approved then it could still be hijacked before that repo is updated? Our pinned dependency versioning should help with that.
This is a good call out. There are alternatives which give finer grained control than `npm build ...`
1. https://github.com/LavaMoat/LavaMoat/tree/main/packages/allow-scripts
2. https://github.com/dominykas/allow-scripts

fallaciousreasoning (approved):
LGTM

thypon:
@AlanBreck they look fine, but unneeded for the moment. We may consider them if we really need the rest of the packages allow script at a later stage.

PR #187, opened 2023-01-27
Generate static JS object reference to the css variables and provide typescript definitions for all JS in generated tokens/ dir

petemill:
Instead of having something seem custom to styled-components, just identify it as JS access to the css variables for the sole reason of intellisense lookups.

The JS object file is now output at `build/css/index.js`. It looks like this:

```ts
import './variables.css'
export default {
  "spacing": {
    "4": "var(--spacing-4)",
    "8": "var(--spacing-8)",
    "12": "var(--spacing-12)",
    "16": "var(--spacing-16)",
    "24": "var(--spacing-24)",
    "32": "var(--spacing-32)",
    "40": "var(--spacing-40)"
  },
  "radius": {
    "1": "var(--radius-1)",
    "2": "var(--radius-2)",
    "4": "var(--radius-4)",
    "8": "var(--radius-8)",
    "12": "var(--radius-12)",
    "16": "var(--radius-16)",
    "full": "var(--radius-full)"
  },
  "color": {
    "light": {
      "text": {
        "primary": "var(--color-light-text-primary)",
        "secondary": "var(--color-light-text-secondary)",
        "interactive": "var(--color-light-text-interactive)",
        "disabled": "var(--color-light-text-disabled)"
      },
      ...
```

This means a consumer can do this:
```ts
import styled from 'styled-components'
import leo from '@brave/leo/tokens/css'

export const StyledWrapper = styled.div`
  flex: 1;
  display: flex;
  position: relative;
  color: ${leo.color.black};
  height: 100%;
`
```

I was also considering we could have a convenience index.js at the root of `@brave/leo/tokens` since it shouldn't interfere with any non-js languages... Just to make the import path, which is sure to be used in every style file of a consuming project, 1 segment shorter...


src/tokens/config.js
```diff
@@ -41,6 +41,11 @@ module.exports = {
             showFileHeader: true,
           },
         },
+        {
+          destination: 'index.ts',
```
fallaciousreasoning, line 45: Maybe this should output a JS file, to be more easily consumable in other apps? As it's just an object, I don't think it'll make any difference to typings

src/tokens/transformation/web/formatCssAsJs.js
```diff
@@ -56,6 +56,7 @@ function formattedVariables (properties) {
 
 module.exports = ({ dictionary, file }) => {
   const themeObject = formattedVariables(dictionary.properties)
-  return fileHeader({ file }) +
-    'export default ' + JSON.stringify(themeObject, null, 2)
+  return fileHeader({ file }) + `
+import './variables.css'
```
fallaciousreasoning, line 60: Hmmm, I don't know if we should do this here. I think consumers will need to import the CSS themselves, as they don't necessarily have support for importing `css` files and I don't think we want to force them to

src/tokens/config.js
fallaciousreasoning: I'm not super opposed to `css/index.js`. Maybe it could be `styles/index.js` and `styles/variables.css` but I don't have strong feelings.

fallaciousreasoning (approved):
LGTM, but with two comments.

src/tokens/transformation/web/formatCssAsJs.js
petemill: I guess that makes sense for vanilla non-bundled apps

AlanBreck: Agreed.

src/tokens/config.js
AlanBreck: Also don't have terribly strong feelings, but lean towards `styles `.

However as I was writing that, I may have convinced myself otherwise because though this is not css, it requires the CSS variables, and therefore isn't really it's own thing. Just a different way of accessing the same thing.

Admittedly an argument could be made that this is also true of Tailwind, but there are other notable differences there like how the variables are built for RGB.

AlanBreck (approved):
Looks good. Weighed in on @fallaciousreasoning's comments.

src/tokens/transformation/web/formatCssAsJs.js
petemill: I wonder if that module should therefore have a check to see if the variables have been added to the document. With the number of individual WebUI in brave-core and our intended other consumers, it could be easy to forget to include a request to that css file when adding a use of this theme, especially since it's not apparent it's using the css variables under the hood. It doesn't seem obvious for all devs to know "I'm using the css theme therefore I need to add a request to css variables." We could add a check something like:
```js
window.requestIdleCallback(() => {
  const value = getComputedStyle(document.documentElement).getPropertyValue('--spacing-4')
  if (!value) {
    console.warn(`Brave Leo css variables have not been added to this document and any use of Leo components or theme variables will be unstyled. Please add a request for '@brave/leo/css/variables.css'.`)
  }
}
```

Likewise, the same is true for the svelte components. Every time someone newly imports any component from Leo, they'll also need to add a request to the css variables. So instead of each svelte component directly importing, maybe they could directly import something like the above (named ~`cssVariableChecker.ts`)?

src/tokens/config.js
petemill: Calling it `styles` seems a bit generic to me, especially since we have `/ios`, `/android`, `/skia` (which maybe should be prefixed cpp as skia has some non-c++ implementations AFIAK), etc... "/styles" could be relevant to any of those languages, whereas this JS object and the css vars can only be used in a platform supporting CSS (e.g. not react native JS either). So to keep with the naming convention of using language or framework, doesn't keeping `/css` seem appropriate?

petemill: > Maybe this should output a JS file,

For some reason I'm not getting intellisense on the object in brave-core when it's a JS file. Will keep experimenting.

AlanBreck: > So to keep with the naming convention of using language or framework, doesn't keeping /css seem appropriate?

I think I've come around and would agree that `css/` is the write place.

src/tokens/transformation/web/formatCssAsJs.js
AlanBreck: I like this. Wonder if instead of something like `--spacing-4`, or anything else that could be found in other libraries, we should add a special var to our css output like `--leo-vars-present: true;`.

src/tokens/config.js
fallaciousreasoning: I don't have strong enough feelings on it to suggest either way. `css` or `styles` is fine :smile:

fallaciousreasoning: Let's just stick with CSS.

src/tokens/transformation/web/formatCssAsJs.js
fallaciousreasoning: Seems like a handy util to me :smile: Probably not critical enough to block this one :smile:

petemill:
I've just pushed the following:
- All .js files in tokens directory are now .ts files
- `npm run transform-tokens` now runs typescript to generate .js and .d.ts files
- Split the css vars lookup js file to multiple exports (1 for each top-level token category).

Tested in brave-core and:
- Get code completion for the properties
- styled-components errors if you give it a token category and not a string value (via the `toString()` property hack)

I think this is good to go if you could have another review @fallaciousreasoning @AlanBreck

src/tokens/config.js
petemill: Kept as .ts in order to successfully generate types (generating .d.ts from .js was providing really strange results...). See PR comment on how we now generate .js files though...

AlanBreck (approved):
Looks good to me. I can't say that I 100% understand the `FAKE_PROPERTY_NAME` bit, but my lack of total understanding isn't a blocker here. :-)

petemill:
> Looks good to me. I can't say that I 100% understand the `FAKE_PROPERTY_NAME` bit, but my lack of total understanding isn't a blocker here. :-)

In summary, it's a way of preventing devs giving a token list to css (e.g. `color.text`) instead of a token (e.g. `color.text.primary`), when they may not be familiar with the property name structure. Since I may forget the reasoning in the future, I'll summarise here more visually than the code comments. Without adding a non-string property, typescript thinks we can give styled-components css the whole tokens object and it will be valid css (because it does accept CSSObject items), i.e. notice no red underline:
image

But when we force the object to not look like a CSSObject by giving it some other non-string property, typescript tells the developer "no".
image

...because the dev really should have used:
image

fallaciousreasoning (approved):
LGTM, with one comment

fallaciousreasoning:
Hey Pete, I think this is pretty good to go once it's rebased!

package.json
```diff
@@ -69,7 +69,7 @@
     "dev": "rollup -c rollup.config.js --watch",
     "update-icons": "node ./src/scripts/update-icons.js",
     "check": "svelte-check --compiler-warnings \"missing-custom-element-compile-options:ignore\"",
-    "transform-tokens": "node ./src/tokens/transformTokens.js",
+    "transform-tokens": "node ./src/tokens/transformTokens.js && tsc -p ./tsconfig-tokens.json",
```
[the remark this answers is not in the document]
petemill: So this build needs to happen after transform-tokens. It isn't really related to the rollup stuff yet, I suppose. We could run from inside the transformTokens.js script. I'm not sure.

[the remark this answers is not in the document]
fallaciousreasoning: Yup, I think you chatted to me about that a bit last week. This seems fine in the case

petemill:
Finally rebased!

petemill:
> I wonder, do you think it's worth linking to your comment above in your code comment? It's much easier to understand 😄
I did that

PR #192, opened 2023-02-01
HTMLElement Props Support

fallaciousreasoning:
This makes it possible to add a whitelisted set of HTMLElement props to the WebComponent. This will allow us to style Leo components with `styled-components` and set common properties, like `style`, `id` and `className` on our Leo components.

It will enable styling like the following:

```ts
import styled from 'styled-components'
import Navdots from '@brave/leo/react/navdots'

const StylishDots = styled(Navdots)`
    --leo-navdots-active-color: hotpink;
    margin-top: 100000px;
`
```


petemill (approved):
The typescript types are a work of art 🎨

AlanBreck:
Wow.

PR #194, opened 2023-02-02
[Component]: Link

### LINKED ISSUE #52: Component: Links
https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?node-id=278%3A12700

image

fallaciousreasoning:
Adds a Link component
Fixes #52

### DIFF
--- a/src/components/link/link.svelte
+++ b/src/components/link/link.svelte
@@ -26,7 +26,7 @@
 </a>
 
 <style lang="scss">
-  .leo-link {
+  a {
     all: unset;
 
     --color: var(--leo-link-color, var(--color-text-interactive));
--- a/src/components/link/link.svelte
+++ b/src/components/link/link.svelte
@@ -1,0 +1,65 @@
+<script lang="ts">
+  import type { SvelteHTMLElements } from 'svelte/elements'
+
+  type ExcludedProps = 'class' | 'href' | 'aria-disabled'
+  type $$Props = Omit<Partial<SvelteHTMLElements['a']>, ExcludedProps> & {
+    isDisabled?: boolean
+    href: string
+  }
+
+  export let href: string
+  export let isDisabled: boolean = false
+</script>
+
+<a
+  rel="noopener"
+  {...$$restProps}
+  {href}
+  class="leo-link"
+  class:disabled={isDisabled}
+  aria-disabled={isDisabled}
+  on:click={(e) => {
+    if (isDisabled) e.preventDefault()
+  }}
+>
+  <slot />
+</a>
+
+<style lang="scss">
+  .leo-link {
+    all: unset;
+
+    --color: var(--leo-link-color, var(--color-text-interactive));
+    --hover-color: var(--leo-link-hover-color, var(--color-primary-60));
+    --visited-color: var(--leo-link-visited-color, var(--color-pink-50));
+    --disabled-color: var(
+      --leo-link-disabled-color,
+      var(--color-text-disabled)
+    );
+    --focus-color: var(--leo-link-focus-color, var(--color));
+    --focus-shadow: var(--leo-link-focus-shadow, var(--effect-focus-state));
+
+    color: var(--color);
+    font: var(--font-text-default-semibold);
+    cursor: pointer;
+    text-decoration: underline;
+
+    &:hover {
+      color: var(--hover-color);
+    }
+
+    &:visited:not(:hover) {
+      color: var(--visited-color);
+    }
+
+    &:focus-visible:not(.disabled) {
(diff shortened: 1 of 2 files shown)


AlanBreck:
Instead of having a link component, can we simply ship default styles anchor tags? I'm not thinking of a scenario when we wouldn't want to use this styling on a Leo based project, and it seems a hassle to import a component for a basic anchor tag.

fallaciousreasoning:
I quite like that idea @AlanBreck - do you think we could just directly style the `a` tag or would we need to add a `.leo-link` class to all links we wanted to style? It'd be kind of nice if we didn't always need to add the class.
```css
a {
    ....
}
```

AlanBreck:
I think we should have sensible defaults that make it as easy for developers to just use standards as possible, so I'd lean towards no special class.

petemill:
We'll have to think about how that integrates with different platforms then. Do we provide a css file for sensible non-component defaults: links, paragraphs, etc. Where do we draw the line? What if a simple component like a link or paragraph or text heading does have a couple variants, like small / large (or a `"wow"` multicolored version)? In that case do we have modifier class names too? Why not do that with Button?

AlanBreck:
> Do we provide a css file...?

I think there could be a strong argument for this. It would allow consumers to "opt-in" to these defaults. We could think of it as a Leo normalizer.

> Where do we draw the line?

While this is admittedly challenging, I think some differences between this and the button are:
- The button styles can be applied to both `<button>` and `<a>`.
- It's easier to assume the default for an anchor tag, whereas for the `Button` component, we don't really want a button to _always_ be primary, unless a unique scenario, but rather should be `primary|secondary|tertiary` based upon purpose.

> The css won't cascade through that boundary unlike the css vars.

This is a bummer. Not really sure what to do here.

> Should we instead just rely on the token css vars and for simple styles just have the consumer always set the font properties themselves? I'd argue no as someone will forget one of the states, like :active

I would agree here.

My concern is that anchor tags are super common, especially in any sort of editorial, longform text. It seems subpar to require that each instance of an anchor tag be our custom `Link` component if the consumer wants the right styles/states. If the consumer is processing Markdown, then links could appear all over, and I wouldn't think we'd want to create a render hook that transforms `[]()` syntax to `<Link href="...">...</Link>`.

fallaciousreasoning:
Maybe we actually need to start providing the CSS separately to the Leo Components, that way consumers could import it directly if they want (and in the case of the Link, we could style the `a` tag).

Our Svelte component would look something like this:
```svelte
<script lang="ts">
....
</script>

<a {...props}><slot/></a>
<link rel="stylesheet" type="text/scss" src="./styles.scss">
```

styles.scss
```scss
a {
 // styles
}
```

And we'd put the compiled scss file into `tokens/css/link.css` (when Svelte compiles, the stylesheet will get dumped into the WebComponent, so that could be a source of duplication).

fallaciousreasoning:
> I think there could be a strong argument for this. It would allow consumers to "opt-in" to these defaults. We could think of it as a Leo normalizer.

I'm pretty in favor of this @AlanBreck, it'd be my preferred solution I think but I worry it would stop people adopting it if they had to do it everywhere at once :/ The component means they can opt in gradually.

> This is a bummer. Not really sure what to do here.

Unfortunately there's not really anything we can do, it's how WebComponents are designed. Individual components can adopt our stylesheet though.

> If the consumer is processing Markdown, then links could appear all over

That's a really good point

AlanBreck:
> Maybe we actually need to start providing the CSS separately to the Leo Components

Will this break the modularization for the CSS that Svelte provides so nicely?

fallaciousreasoning:
I was actually just talking to @petemill about this - I think we'll end up with 3 states :laughing: 
1. CSS in the Svelte components. This is compiled by Svelte and nicely modularized.
2. CSS in the Web Components. This is compiled (but only from SCSS to CSS or whatever). It's scoped naturally by the shadow DOM.
3. CSS. This is compiled (from SCSS to CSS) and placed in `tokens/css/<component>.css`. It is unscoped, using it applies to everything in the same Document/Shadow DOM

src/components/link/link.svelte
```diff
@@ -0,0 +1,65 @@
...
+  aria-disabled={isDisabled}
+  on:click={(e) => {
+    if (isDisabled) e.preventDefault()
+  }}
+>
+  <slot />
+</a>
+
+<style lang="scss">
+  .leo-link {
```
petemill, line 29: Change to `a` for future-proof on css extraction

petemill (approved):

PR #196, opened 2023-02-07
Cleanup deployed prs

kim0:
Resolves https://github.com/brave/devops/issues/8849

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -4,9 +4,6 @@ on:
   schedule:
     - cron: '30 5 * * 1'
   workflow_dispatch:
-  push:
-    branches:
-      - 'aka/cleanup_prs'
 
 jobs:
   cleanup:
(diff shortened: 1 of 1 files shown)


mihaiplesa (approved):

.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,19 @@
+name: cleanup
+
+on:
+  schedule:
+    - cron: '30 5 * * 1'
+  workflow_dispatch:
+  push:
+    branches:
+      - 'aka/cleanup_prs'
```
AlanBreck, line 9: Does this need to be removed prior to merge?

AlanBreck (approved):

fallaciousreasoning (approved):
LGTM - out of interest we can't set this up to run when branches merge?

kim0:
@fallaciousreasoning That can certainly work too, I opted for this approach bec we have to ping GH about every open PR which can get kinda expensive on busy repos. This is an approach I used before elsewhere and worked well so I went with it.

.github/workflows/cleanup.yml
kim0: Correct. Removed

mihaiplesa (approved):

PR #197, opened 2023-02-07
New design tokens structure

aguscruiz:
Added universal / ios / android .tokens.json files

Also, renamed font styles in universal, from "Text" to "Primary" and from "LongText" to "Secondary"

### DIFF
--- a/src/tokens/android.tokens.json
+++ b/src/tokens/android.tokens.json
@@ -1,0 +1,1198 @@
+{
+  "font": {
+    "🤖 android": {
+      "heading": {
+        "display1": {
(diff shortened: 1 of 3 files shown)


fallaciousreasoning:
@aguscruiz I've deleted the `design-tokens.tokens.json` as it's been replaced by the new files. I think it'd be best if @petemill or @AlanBreck had a look though as I'm pretty hazy on the `design-tokens` stuff.

petemill (approved):

petemill (approved):
We compared the output and the only changes are the text variables names - thanks @aguscruiz !

PR #200, opened 2023-02-09
adds package.json to tokens/tailwind

AlanBreck:
The usual way to import a Tailwind preset is to use `require` like so:

```js
/** @type {import('tailwindcss').Config} */
module.exports = {
	content: ["./**/*.html"],
	presets: [
		require("@brave/leo/tokens/tailwind")
	],
	theme: {},
	plugins: [],
};
```

Since the nearest `package.json` uses `"type": "module"`, this causes the `require` statement to fail. As such, I'm proposing we just include a basic `package.json` file to the `tokens/tailwind` folder to counteract this.


fallaciousreasoning (approved):
Seems fine to me - it's kinda madness the node ecosystem makes us do this :laughing: - looks like you need to run `npm run format` though

PR #201, opened 2023-02-09
Progress Improvements

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/202
- Makes the progress ring resizable (while maintaining the ability to set the stroke width & and CSS variables)
- Split the `progress` story into a `progressRing` and `progressBar` story
- Update the stories to list out the CSS variables
- Type the mode, so it's obvious what the options are


AlanBreck (approved):
This looks great. Thank you!

PR #212, opened 2023-02-12
Fix build & tests

fallaciousreasoning:
- Fixes breaking change from `svelte2tsx` bump which changed the output folder of the type definitions
- Remove `jest` from `package.json` as we're using `ts-jest`
- Move `ts-jest` from 'dependencies` to `devDependencies`
- Fix broken import in test

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,6 @@
         "svelte-preprocess": "5.0.1",
         "svelte2tsx": "0.6.1",
         "tailwindcss": "3.2.6",
-        "ts-jest": "29.0.5",
         "tslib": "2.5.0"
       },
       "devDependencies": {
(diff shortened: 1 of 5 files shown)


AlanBreck (approved):
LGTM

Sorry about the regressions due to the `svelte-check` update.

fallaciousreasoning:
Thanks @AlanBreck - no stress for bumping the version, it was nearly impossible to tell when the build fails because I think the check was disappearing from the list of checks, rather than showing as red :laughing: - I asked IT support to turn on branch protection so we can't merge if the build task fails, so we'll see if that helps :laughing: 

It was also good 'cause I realised my tests were broken (and had been for a while - they were trying to import something from the wrong folder). I guess we aren't running them either?

PR #214, opened 2023-02-13
forces resolution for axios

AlanBreck:
We have multiple dependapot alerts due to an [outdated version of axios which `figma-js` depends upon](https://github.com/jemgold/figma-js/blob/main/package.json#L57).

I thought about submitting a PR to the `figma-js` project, but got errors as soon as I tried `npm install` due to conflicts with `tslint`, which has apparently been deprecated since 2019.

This seems to be the path of least resistance.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "type": "module",
   "dependencies": {
-    "@ctrl/tinycolor": "3.5.0",
+    "@ctrl/tinycolor": "3.5.1",
     "@tsconfig/svelte": "3.0.0",
     "lodash.camelcase": "4.3.0",
     "lodash.merge": "4.6.2",
@@ -38,7 +38,6 @@
     "css-loader": "6.7.3",
     "css-tree": "2.3.1",
     "figma-api-exporter": "github:brave/figma-api-exporter#775a17babce35ed165936142caf5c5dca3490395",
-    "ts-jest": "29.0.5",
     "jsdom": "21.1.0",
     "postcss": "8.4.21",
     "postcss-js": "4.0.1",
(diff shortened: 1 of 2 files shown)


fallaciousreasoning:
Looks like installing scripts failed on CI?

package.json
```diff
@@ -77,7 +77,11 @@
     "transform:android": "node ./examples/android/build.js",
     "transform:ios": "node ./examples/ios/build.js",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "preinstall": "npx npm-force-resolutions"
```
fallaciousreasoning, line 81: I wonder if this is a potential security issue? It'll run whichever the most recent version is so if they release a malicious version it will get run on CI.

Instead, we could add it as a `devDependency` and change the script to
```suggestion
    "preinstall": "npm-force-resolutions"
```

fallaciousreasoning: Also, this has completely changed the `package-lock.json` file. Do we know if it's going to be stable across runs?

AlanBreck: I've just replaced this with [`overrides` feature of `npm`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides).

fallaciousreasoning (approved):

AlanBreck:
Ignoring the following Socket issues due to a) presence of `--ignore-scripts` and b) previously handling for Jest bin conflicts.

@SocketSecurity ignore core-js-pure@3.23.2
@SocketSecurity ignore jest@29.4.2
@SocketSecurity ignore jest-cli@27.5.1
@SocketSecurity ignore jest-cli@29.4.2

PR #216, opened 2023-02-15
Tokens: add "leo-" prefix to css variable names

petemill:
When reading Leo component source css, or consuming apps' components, it's difficult to differentiate local css variables that a component might use for different states, and leo variables. For example a component may locally set and read `--color-text` whereas Leo contains `--color-text-primary` which that component may also use. 

Not only does this become difficult to quickly read and know which variables are local and which are global, but it can cause unintentional variable name conflicts and overrides, creating visual issues when a component has descendants which read from the unintentionally overridden variables.

PR also contains:
- [remove unused -light- and dark- name prefix stripping. It wasn't used. There are no changes in the variables.css output](https://github.com/brave/leo/commit/a005f5a3997808ab8fe517ad13998b6eba0f0283) 
- [Tokens: Refactor css property name string construction so that it's shared between css and js output](https://github.com/brave/leo/commit/6030727f2e33410e2bb0be7673cb9c0798895334)


src/tokens/transformation/web/createPropertyFormatter.js
```diff
@@ -19,40 +19,7 @@ const defaultFormatting = {
...
- * @param {Object} options.formatting - Custom formatting properties that define parts of a declaration line in code. The configurable strings are: prefix, indentation, separator, suffix, and commentStyle. Those are used to generate a line like this: `${indentation}${prefix}${prop.name}${separator} ${prop.value}${suffix}`
- * @returns {Function}
- */
-function createPropertyFormatter({
-  outputReferences,
-  dictionary,
-  format,
-  formatting = {}
-}) {
+function createPropertyNameFormatter(format, formatting = {}) {
```
petemill, line 22: `createPropertyNameFormatter` extracted from `createPropertyFormatter` so that it can be used by `formatCssAsJs` instead of making this change twice.

AlanBreck:
Are there any changes in the Tailwind output?

fallaciousreasoning:
We need to update all the components too - none of the CSS variables they're using exist now

Button
image

Collapse
image

Label
image

Link
image

Navdots
image

Progress Bar
image

Progress Ring
image

Toggle
image

petemill:
@fallaciousreasoning oh of course, I was so preoccupied by the tokens output that I ignored that!

petemill:
Components updated

petemill:
@AlanBreck generated `tokens/tailwind` directory is unchanged

```
~/Development/Brave/leo main*
❯ diff -r ./.tailwind tokens/tailwind
```
no diff where `.tailwind` is the output from this branch and `tokens/tailwind` is the output from `main` branch

petemill:
Storybook looks indentical now too

AlanBreck (commented):
Sorry to be a fly in the ointment (and I'm happy to let this go for now if best), but the `--leo` prefix for both token variables and component variables seems to make things less then clear, and still challenging to know what's a token, and what's specific to the component:

```css
--color: var(--leo-link-color, var(--leo-color-text-interactive));
--hover-color: var(--leo-link-hover-color, var(--leo-color-primary-60));
--visited-color: var(--leo-link-visited-color, var(--leo-color-pink-50));
```

Brad Frost's Atomic Design Patterns loosely uses the hierarchy of atoms, molecules, etc. to distinguish between different levels of abstraction. Could we do something similar with Leo to distinguish between tokens and components? `--leo-token` and `--leo-component` are obviously clunky, but is there some other model we could adopt that is (a) distinct enough from other systems to be a namespace of sorts, and (b) succinct enough that we could omit `--leo` in favor of the shorter prefix that points to the variable's purpose?

I don't really have specific ideas, so I apologize if the above seems really vague and squishy.

petemill:
@AlanBreck I have similar thoughts, but as I was making these changes I still think that 
- Adding the `--leo-` prefix has made it much easier (for me) to read these components inside Leo source code. There are so many local variables with the prefix, e.g. `--color-x` that this is still a big improvement on readability. More so than the remaining `leo-[component-name]-variable` instances.
- I quickly got used to seeing the difference `leo-[component-name]` vs `leo-color` or `leo-font` and more easily distinguishing them than before. It's at least possible, given you know the name of the component that you're looking at the source code of.
- Adding the global prefix improves readability in consuming apps outside of Leo. The component variables will show up much less there and always as setting them, not reading their values, i.e. on the left side and not the right side. Whereas the global variables should always be read as values, appear on the right side and really not appear on the left side.

Can we do this iteratively? I personally can't imagine a better name for the global variables than simply `--leo-`. It is concise and makes sense for consumers. So if we come up with a better name for component-specific style property overrides then perhaps we can add that in a subsequent PR?

AlanBreck (approved):

fallaciousreasoning (approved):
LGTM 2

PR #217, opened 2023-02-15
Update Stories with CSS props

fallaciousreasoning:


AlanBreck (approved):
LGTM!

Minor nit: it appears that storybook is treating `isOpen` in the `Collapse` component as `any`, but it should be a `boolean`, right?
Screenshot 2023-02-16 at 10 33 04 AM

fallaciousreasoning:
So it'd be ideal if it didn't show that at all :laughing: if you set it it breaks clicking to open/close the collapse

Issue #220, opened 2023-02-20
Wrong flag icons


kkarikos:
There is wrong flag icons for:
- United Kingdom (GB) flag is flag of Wales

Missing:
- Netherlands (NL) (probably due to some naming issue)

PR #221, opened 2023-02-20
Update icons

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/220


src/scripts/update-icons.js
```diff
@@ -23,7 +23,7 @@ const getFlag = (iconName) => {
     return `${match[1]}`
   }
 
-  const countryRegex = /Country=(\w\w).*\.svg/
+  const countryRegex = /Country=(\w\w(-\w\w\w)?).*\.svg/
```
fallaciousreasoning, line 26: This was a bug where a bunch of the UK icons weren't being parsed correctly

AlanBreck (approved):

fallaciousreasoning:
Leaving this open for now - I think some of the icons need to be updated in Figma

aguscruiz:
@fallaciousreasoning I already renamed the ones with GB- to remove that prefix, so I think we're good to go for now

PR #222, opened 2023-02-22
Add Skia Icons to output

fallaciousreasoning:
This will let us use the Leo Icons from inside Views

Fixes https://github.com/brave/leo/issues/223


src/scripts/gen-skia-icons.js
```diff
@@ -0,0 +1,21 @@
...
+const ICONS_FOLDER = "icons/"
+const OUTPUT_FOLDER = "icons-skia/"
+
+import('skiafy').then(async ({ default: skiafy }) => {
+    await fs.mkdir(OUTPUT_FOLDER, { recursive: true })
+
+    const skiafyPromise = await fs.readdir(ICONS_FOLDER)
+        .then(icons => icons.filter(i => i.endsWith('.svg')))
+        .then(icons => icons.map(async icon => {
+            const svgContent = await fs.readFile(path.join(ICONS_FOLDER, icon), { encoding: 'utf8' })
```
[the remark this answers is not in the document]
thypon: `.replace('-', '_') + '.icon')` is probably wrong here since we are replacing both the path basename and the OUTPUT folder dashes (`-`).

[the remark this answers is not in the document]
fallaciousreasoning: n/a

src/scripts/gen-skia-icons.js
```diff
@@ -0,0 +1,30 @@
...
+
+import('skiafy').then(async ({ default: skiafy }) => {
+  await fs.mkdir(OUTPUT_FOLDER, { recursive: true })
+
+  const skiafyPromise = await fs
+    .readdir(ICONS_FOLDER)
+    .then((icons) => icons.filter((i) => i.endsWith('.svg')))
+    .then((icons) =>
+      icons.map(async (icon) => {
+        const svgContent = await fs.readFile(path.join(ICONS_FOLDER, icon), {
```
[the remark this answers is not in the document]
fallaciousreasoning: n/a

thypon (approved):

src/scripts/gen-skia-icons.js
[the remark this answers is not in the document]
fallaciousreasoning: `icon` should just be the name of the icon file (i.e. `arrow-left.svg`). It gets `path.join`'d with the output folder on line 23 :smile:

package.json
```diff
@@ -72,8 +73,9 @@
     "dev": "rollup -c rollup.config.js --watch",
     "update-icons": "node ./src/scripts/update-icons.js",
     "check": "svelte-check --compiler-warnings \"missing-custom-element-compile-options:ignore\"",
-    "transform-tokens": "node ./src/tokens/transformTokens.js && tsc -p ./tsconfig-tokens.json",
+    "transform-tokens": "node ./src/tokens/transformTokens.js && tsc -p ./tsconfig-tokens.json && npm run transform:icons",
```
petemill, line 76: hmm, these aren't tokens, so I'm wondering if we just leave it as a separate command, but bundle with  `npm run build`. Then there's still a way just to update the tokens.

AlanBreck: This actually could probably move to an action for Skia instead of being a separate npm script. Cf: https://github.com/brave/leo/tree/main/src/tokens/transformation/tailwind/index.js#L70-L74

package.json
```diff
@@ -72,8 +73,9 @@
...
     "test": "jest tests --coverage=false",
+    "transform:icons": "node ./src/scripts/gen-skia-icons.js",
     "transform:android": "node ./examples/android/build.js",
     "transform:ios": "node ./examples/ios/build.js",
```
AlanBreck, line 80: Unrelated to this PR, can we remove these?

package.json
```diff
@@ -47,6 +47,7 @@
     "rollup-plugin-svelte": "7.1.2",
     "sass": "1.58.0",
     "sass-loader": "13.2.0",
+    "skiafy": "github:fallaciousreasoning/skiafy",
```
fallaciousreasoning, line 50: Can't merge until we fork this into Brave

petemill, line 50: So this `skiafy` that we forked[0] seems to be a conversion of the original skia repo[1] that works in node. It seems like it hasn't been maintained in 4 years. I suggest we take the current skiafy script and make the same node-compatibility changes as just put it in leo, as this is the only place we're using it at Brave and we don't need the added complexity of another fork that we won't look at for years.

[0] https://github.com/zhsoft88/skiafy
[1] https://github.com/evanstade/skiafy

fallaciousreasoning: Moved in to Leo :smile:

fallaciousreasoning: Removed dep.

package.json
fallaciousreasoning: Moved to it's own script, which get run at `build`

package.json
fallaciousreasoning: https://github.com/brave/leo/pull/224

AlanBreck (approved):

PR #225, opened 2023-02-27
Improves Skia output to make it easier to use in Brave Core

fallaciousreasoning:
Questions:
~1. Should I be doing substitution to make the dark colors automatically work (i.e. `kColorPrimary40` is `primary40` in light and `primary60` in dark).~ This is handled in Figma
2. Should I depend on the Chromium `PreferredColorScheme` type instead of introducing `leo::Theme::{kDark,kLight}`

### DIFF
--- a/src/tokens/transformation/skia/templates/colors.h.template
+++ b/src/tokens/transformation/skia/templates/colors.h.template
@@ -16,3 +16,27 @@ namespace leo::dark {
   return `constexpr SkColor ${prop.name} = SkColorSetRGB(${prop.value})`;
 }).map(s => `${s};`).join('\n') %>
 } // namespace leo::dark
+
+namespace leo {
+
+enum Theme {
+  kLight,
+  kDark
+};
+
+enum Color {
+<%= groupedTokens.light.allTokens.map(prop => `  ${prop.name}`).join(',\n') %>
+};
+
+SkColor GetColor(Color color, Theme theme) {
+  switch (color) {
+<%= groupedTokens.light.allTokens.map(prop => `    case ${prop.name}:
+      return theme == Theme::kLight
+        ? leo::light::${prop.name}
+        : leo::dark::${prop.name};`).join('\n') %>
+    default:
+      return SK_ColorRED;
+  }
+}
+
+} // namespace leo
--- a/src/tokens/transformation/skia/templates/colors.h.template
+++ b/src/tokens/transformation/skia/templates/colors.h.template
@@ -19,16 +19,16 @@ namespace leo::dark {
 
 namespace leo {
 
-enum Theme {
+enum class Theme {
   kLight,
   kDark
 };
 
-enum Color {
+enum class Color {
 <%= groupedTokens.light.allTokens.map(prop => `  ${prop.name}`).join(',\n') %>
 };
 
-SkColor GetColor(Color color, Theme theme) {
+constexpr SkColor GetColor(Color color, Theme theme) {
   switch (color) {
 <%= groupedTokens.light.allTokens.map(prop => `    case ${prop.name}:
       return theme == Theme::kLight
--- a/src/tokens/config.js
+++ b/src/tokens/config.js
@@ -60,7 +60,7 @@ module.exports = {
       buildPath: 'tokens/skia/',
       files: [
         {
-          destination: 'Colors.h',
+          destination: 'colors.h',
           format: 'skia/colors.h',
           filter: {
             type: 'color'


petemill (approved):
> Should I be doing substitution to make the dark colors automatically work (i.e. kColorPrimary40 is primary40 in light and primary60 in dark).

The way you currently have it in this branch is the same way we're doing it in css, so I'm not sure I follow which `primary60` (light or dark) would be the value for `kColorPrimary40` in dark? I probably just don't understand the question.

> Should I depend on the Chromium PreferredColorScheme type instead of introducing leo::Theme::{kDark,kLight}

Probably the way you've done it here is fine, especially if there'll be some kind of central function in a useful Views / Color-related chromium / brave-core class instance that will wrap this and provide the correct theme?

In general though, I think this is great.

fallaciousreasoning:
I think more likely, I just don't understand what I'm asking :laughing: I thought we did some cleverness in CSS so that the colors Just Work(tm) but maybe the UX folks do that (in which case, wonderful!)

src/tokens/transformation/skia/templates/colors.h.template
```diff
@@ -16,3 +16,27 @@ namespace leo::dark {
   return `constexpr SkColor ${prop.name} = SkColorSetRGB(${prop.value})`;
 }).map(s => `${s};`).join('\n') %>
 } // namespace leo::dark
+
+namespace leo {
+
+enum Theme {
```
sangwoo108, line 22: I think `emum class` would be better as there's multiple enums in this scope.

src/tokens/transformation/skia/templates/colors.h.template
```diff
@@ -16,3 +16,27 @@ namespace leo::dark {
...
+  kLight,
+  kDark
+};
+
+enum Color {
```
sangwoo108, line 27: How about using E_CPONLY(), which is used by brave_color_ids.h.

src/tokens/transformation/skia/templates/colors.h.template
```diff
@@ -16,3 +16,27 @@ namespace leo::dark {
...
+<%= groupedTokens.light.allTokens.map(prop => `  ${prop.name}`).join(',\n') %>
+};
+
+SkColor GetColor(Color color, Theme theme) {
```
sangwoo108, line 31: I think it'd be good to migrate this into color provider facilities. Then, Given that, we could get colors from views like so,

```
void OnThemeChnaged() {
  GetColorProvider()->GetColor(kLeoFooColor);
}

```

But this would be available with changes from Brave-core

sangwoo108: Making this function `constexpr function` could be better.

src/tokens/transformation/skia/templates/colors.h.template
```diff
@@ -16,3 +16,27 @@ namespace leo::dark {
...
+  switch (color) {
+<%= groupedTokens.light.allTokens.map(prop => `    case ${prop.name}:
+      return theme == Theme::kLight
+        ? leo::light::${prop.name}
+        : leo::dark::${prop.name};`).join('\n') %>
+    default:
+      return SK_ColorRED;
```
sangwoo108, line 38: optional: gfx::kPlacholderColor could be used for things like this - but effectively same 😄

fallaciousreasoning: It'd probably be better, but I'm hesitant to add any more Brave specific dependencies than strictly necessary :laughing: I might leave this one for now.

src/tokens/transformation/skia/templates/colors.h.template
fallaciousreasoning: Sweet!

src/tokens/transformation/skia/templates/colors.h.template
fallaciousreasoning: Hmm, that looks interesting - it says it provides a string name for the color, what exactly is the purpose? Is it for when you enable devtools for views?

I think I'd feel (marginally) better leaving it out for now, and coming back and adding it when we need it. What do you think?

(I haven't encountered it before, I'm happy to add it if you think it'll be useful though)

src/tokens/transformation/skia/templates/colors.h.template
fallaciousreasoning: C++ can do some mad things :laughing: That's quite cool

PR #228, opened 2023-03-03
Updated universal / ios / android tokens

aguscruiz:
Changed light/container/highlight value
Also made native adjustments for font values in iOS and Android.

Btw @AlanBreck the token file I had "design-tokens.tokens.json" should be replaced by "universal.tokens.json" I renamed it to match the library name. Hope that's ok

### DIFF
--- a/src/tokens/universal.tokens.json
+++ b/src/tokens/universal.tokens.json
@@ -366,7 +366,7 @@
         "highlight": {
           "description": "",
           "type": "color",
-          "value": "#f4f6f8ff",
+          "value": "#eff2f5ff",
           "blendMode": "normal",
           "extensions": {
             "org.lukasoppermann.figmaDesignTokens": {
@@ -378,7 +378,7 @@
         "interactive-background": {
           "description": "",
           "type": "color",
-          "value": "#f5f4feff",
+          "value": "#f2f1feff",
           "blendMode": "normal",
           "extensions": {
             "org.lukasoppermann.figmaDesignTokens": {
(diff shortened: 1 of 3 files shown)


src/tokens/universal.tokens.json
```diff
@@ -3097,6 +3097,30 @@
...
+              }
+            ]
+          },
+          "extensions": {
+            "org.lukasoppermann.figmaDesignTokens": {
+              "styleId": "S:7d40c65bebc9cad2488922f2c2b931edb969a481,",
+              "exportKey": "gradient"
+            }
+          }
+        },
```
AlanBreck, line 3123: Just throwing this out there as a question: should `rewards-panel` be a part of the design system, or is that getting too project specific? If it's not meant to be used elsewhere, it seems the rewards panel code base should own this, and if it is meant to be used elsewhere, it seems it should be named to something more generic.

fallaciousreasoning: :+1: to this - I think it'd be ideal if we could avoid adding project specific properties to Leo - if we have some legitimate additions, they can probably be added at a project specific level.

AlanBreck: cc @aguscruiz

AlanBreck (approved):
LGTM!

aguscruiz:
> I think we can merge this when the conflicts are resolved - are you okay to do that @aguscruiz?

I'm good! I just pushed a not significant update for some gray values, but nothing more.

Is there anything I can do to fix the conflict?

PR #234, opened 2023-03-10
[Dependencies]: Point to upstream Svelte

fallaciousreasoning:
Once we upgrade Leo consumers we should be able to delete our fork.

https://github.com/sveltejs/svelte/issues/8134#issuecomment-1463047453


AlanBreck (approved):
Resounding yes!

PR #239, opened 2023-03-17
Svelte doesn't allow Web Components which contain Svelte components.

fallaciousreasoning:
Fix https://github.com/brave/leo/issues/245
Fix https://github.com/brave/leo/issues/244

Unfortunately, this means we need to write our own wrapper. On the bright side, Svelte's web components are **not** well supported and have a number of bugs including:
- Nested component slots conflict with parent component slots
- Nested components styles are not applied

This means we have a bit more room to fix bugs / make changes to how things work. On the other hand, this might make life harder if Svelte changes things down the tracks. We do get a few wins from this:
1. I suspect we'll have a smaller bundle size because more code is shared
2. We can turn child slots on & off! (in Svelte an `{#if false}` is treated as slotted content).
3. We can use advanced Svelte behavior (like bind) inside Leo, although it won't be supported via the web component API. With Svelte's built in Web Components, this does not work.
 
~~This PR is not quite ready yet, as I need to decide how we want the output to look, and I need to do a bit of testing of our React wrapper.~~
- [x] ~~React wrapper testing~~
- [x] ~~Removing event listeners~~

### DIFF
--- a/src/components/svelte-react.ts
+++ b/src/components/svelte-react.ts
@@ -5,7 +5,8 @@ import {
   forwardRef,
   useCallback,
   type ForwardedRef,
-  type PropsWithChildren
+  type PropsWithChildren,
+  useState
 } from 'react'
 import type { SvelteComponentTyped } from 'svelte'
 
@@ -32,13 +33,14 @@ export type ReactProps<Props, Events> = Props & {
 } & {
   ref?: ForwardedRef<Partial<Props & HTMLElement> | undefined>
 } & {
-  // Note: The div here isn't important because all props in intrinsicProps are
-  // available on all elements. We just want to make sure we have the correct
-  // React name/value for them.
-  [P in IntrinsicProps]?: JSX.IntrinsicElements['div'][P]
-}
+    // Note: The div here isn't important because all props in intrinsicProps are
+    // available on all elements. We just want to make sure we have the correct
+    // React name/value for them.
+    [P in IntrinsicProps]?: JSX.IntrinsicElements['div'][P]
+  }
 
-const useEventHandlers = (el: HTMLElement | undefined, props: any) => {
+const useEventHandlers = (props: any) => {
+  const [el, setEl] = useState<HTMLElement>(null)
   const lastValue = useRef<{ [key: string]: (...args: any[]) => any }>({})
 
   // Handle updating event listeners when props change
@@ -64,15 +66,22 @@ const useEventHandlers = (el: HTMLElement | undefined, props: any) => {
       el.removeEventListener(removed, lastValue.current[removed])
       delete lastValue.current[removed]
     }
-  }, [props])
+  }, [props, el])
 
-  // Handle cleaning up event listeners when destroying component
-  useEffect(() => () => {
-    if (!el) return
-    for (const [event, listener] of Object.entries(lastValue.current)) {
-      el.removeEventListener(event, listener)
-    } 
+  const setElement = useCallback((el: HTMLElement | null) => {
+    lastValue.current = {}
+    setEl(oldValue => {
+      // Cleanup
+      for (const [event, listener] of Object.entries(lastValue.current)) {
+        oldValue.removeEventListener(event, listener)
+      }
+      return el
+    })
   }, [])
+
+  return {
+    setElement
+  }
 }
 
 /**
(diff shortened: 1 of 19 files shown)


AlanBreck (approved):
Looks good to me! I'm not well versed in web-components, so I'd definitely defer to @petemill for a more definitive review.

src/components/svelte-web.ts
```diff
@@ -0,0 +1,226 @@
...
+          enumerable: true,
+          get() {
+            const contextIndex = c.$$.props[prop]
+            return this.component.$$.ctx[contextIndex]
+          },
+          set(value) {
+            if (reflectToAttributes.has(typeof value)) {
+              this.setAttribute(prop, value)
+            }
+            this.component.$$set({ [prop]: value })
```
AlanBreck, line 194: Why `.$$set` instead of [`.$set` as documented here](https://svelte.dev/docs#run-time-client-side-component-api-$set)?

src/components/svelte-web.ts
```diff
@@ -0,0 +1,226 @@
...
+          .reduce((prev, [key, value]) => ({ ...prev, [key]: value }), {})
+
+        // If the component already exists, destroy it. This is,
+        // unfortunately, necessary as there is no way to update slotted
+        // content in the output Svelte compiles to. This is a problem
+        // even when not doing crazy things:
+        // https://github.com/sveltejs/svelte/issues/5312
+        if (this.component) {
+          this.component.$destroy()
+        }
```
petemill, line 139: I wonder what side effects this can produce. I guess there shouldn't really be any for Leo's simple components that shouldn't do much on 'mount', or maintain state. Although we're about to introduce AlertCentre which seems to! (not via slots though thankfully)

src/components/svelte-react.ts
```diff
@@ -67,6 +45,58 @@ export type ReactProps<Props, Events> = Props & {
   [P in IntrinsicProps]?: JSX.IntrinsicElements['div'][P]
 }
 
+const useEventHandlers = (props: any) => {
+  const [el, setEl] = useState<HTMLElement>()
+  const lastValue = useRef<{ [key: string]: (...args: any[]) => any }>({})
```
petemill, line 50: Would be good to comment what we're storing in `lastValue` - last value of what? Maybe `currentEventHandlers`?

src/components/svelte-react.ts
```diff
@@ -67,6 +45,58 @@ export type ReactProps<Props, Events> = Props & {
...
+
+  // Handle updating event listeners when props change
+  useEffect(() => {
+    if (!el) return
+    for (const [key, listener] of Object.entries(props) as [
+      string,
+      (...args: any[]) => any
+    ][]) {
+      const match = eventRegex.exec(key)
```
petemill, line 59: Perhaps an optimization so that we're only running the regex once per prop key? Some kind of Map of keyname to boolean representing if it's an event prop, or undefined if we haven't checked the regex yet?

petemill (approved):
This is really great. Seems like you've thought of every edge case and understood the svelte internals very well. It does seem to solve the couple of issues we're experiencing with the react binding. Perhaps as a follow-up we can add more comments to the uses of svelte's internal API so that if we have to fix changes there in the future it's quicker for us to see why we're using that API feature.

src/components/svelte-web.ts
```diff
@@ -0,0 +1,226 @@
...
+export default function registerWebComponent(
+  component: any,
+  { name, mode }: Options
+) {
+  // Create & mount a dummy component. We use this to work out what props are
+  // available and generate a list of available properties.
+  const c = new component({ target: document.createElement('div') })
+
+  // The names of all properties on our Svelte component.
+  const props = Object.keys(c.$$.props)
```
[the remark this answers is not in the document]
fallaciousreasoning: Yeah, that'd be pretty neat :D Somtimes I wonder if we're going to end up with our own UI framework :laughing:

src/components/svelte-web.ts
fallaciousreasoning: Definitely if a component is doing something on mount we're going to get some interesting side effects. Unfortunately I don't really see a way around this for now.

We should be able to maintain state between mounts without too much difficulty but yeah, side effects are going to be a bit of a footgun.

src/components/svelte-react.ts
fallaciousreasoning: Do you mind if we just leave this for now and just see if it's slow?

src/components/svelte-web.ts
fallaciousreasoning: Because when I did `Object.entries` on a Svelte component `$$set` showed up and `$set` did not :laughing:. I've swapped to using this in a new PR, as it definitely seems like the (more) blessed path.

PR #240, opened 2023-03-17
The Button, when used as a web component, should be inline-block by default

petemill:
Allowing it to grow inline by default, or be stretched to a specific width via flexbox or a specific width, etc


fallaciousreasoning (approved):
Lgtm

PR #242, opened 2023-03-21
In lieu of tests, use react example app to ensure there are no type errors

petemill:
When reviewing https://github.com/brave/leo/pull/239 I had a typescript error in the react example app I needed to solve. This runs that build on CI so we can see errors earlier.

Actual react component tests would be better, but this works for now.


examples/example-ui-react/.gitignore
AlanBreck, line 1: Shouldn't we still ignore `node_modules`?

petemill: I'm pretty sure that child `.gitignore`s are additive - the parent `.gitignore` rules still apply. And it's working - it's not prompting me to commit the `node_modules` dir in this directory.

PR #246, opened 2023-03-23
feat(icon): add & update props

josheleonard:
Added suggestions to `name` prop type
Added `size` prop
Added `color` & `darkModeColor` prop

### DIFF
--- a/icons/meta.ts
+++ b/icons/meta.ts
@@ -1778,4 +1778,4 @@ export default {
     'zilliqa-color': 'zilliqa-color',
     'zrx-color': 'zrx-color'
   }
-}
+} as const
--- a/src/components/icon/icon.svelte
+++ b/src/components/icon/icon.svelte
@@ -4,13 +4,23 @@
 </script>
 
 <script lang="ts">
-  export let name: string = undefined
+  import type { IconNames, StringWithAutocomplete } from '../types'
+
+  export let name: StringWithAutocomplete<IconNames> = undefined
   export let forceColor: boolean = false
+  export let size: string = 'unset'
+  export let color: string | undefined = 'unset'
+  export let darkModeColor: string | undefined = 'unset'
   let hasColor =
     name?.endsWith('-color') || name?.startsWith('Country=') || forceColor
 </script>
 
-<div class="leoIcon">
+<div
+  class="leoIcon"
+  style:--size-override={size}
+  style:--color-override={color}
+  style:--dark-mode-color-override={darkModeColor}
+>
   <slot>
     {#if name}
       <div
@@ -24,11 +34,13 @@
 
 <style lang="scss">
   .leoIcon {
-    --icon-width: var(--leo-icon-size, 24px);
-    --icon-height: var(--leo-icon-size, 24px);
+    --icon-size: var(--size-override, var(--leo-icon-size, 24px));
 
-    width: var(--icon-width);
-    height: var(--icon-height);
+    --icon-light-mode-color: var(--color-override, currentColor);
+    --icon-dark-mode-color: var(--dark-mode-color-override, currentColor);
+
+    width: var(--icon-size);
+    height: var(--icon-size);
 
     & .icon,
     svg {
@@ -39,10 +51,14 @@
     /* Non color icons are set via a mask-image, so we can change the color
      * by setting the background */
     & .icon:not(.color) {
-      background: currentColor;
+      background: var(--icon-light-mode-color);
       -webkit-mask-image: var(--icon-url);
       mask-image: var(--icon-url);
       mask-repeat: no-repeat;
+
+      @media (prefers-color-scheme: dark) {
+        background: var(--icon-dark-mode-color);
+      }
     }
 
     /* Color icons need to be set via a background rather than a mask so we
--- a/src/components/types.ts
+++ b/src/components/types.ts
@@ -1,0 +1,10 @@
+// Copyright (c) 2023 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import type meta from '../../icons/meta'
+
+export type StringWithAutocomplete<T> = T | (string & Record<never, never>)
+
+export type IconNames = keyof (typeof meta)['icons']
--- a/src/components/icon/icon.stories.svelte
+++ b/src/components/icon/icon.stories.svelte
@@ -37,7 +37,7 @@
 
 <Template let:args>
   <div style={`color: ${args.color}; --leo-icon-size: ${args.size}px;`}>
-    <Icon {...args} />
+    <Icon name={args.name} />
   </div>
 </Template>
 
@@ -45,10 +45,10 @@
 
 <Story name="Custom Icon" let:args>
   <div style={`color: ${args.color}; --leo-icon-size: ${args.size}px;`}>
-    This icon isn't rendered from our libray - instead an SVG is passed in via a
-    slot. This allows consumers of the libary to import their own SVGs and use
-    them directly inside an icon.
-    <Icon {...args}>
+    This icon isn't rendered from our library - instead an SVG is passed in via
+    a slot. This allows consumers of the library to import their own SVGs and
+    use them directly inside an icon.
+    <Icon name={args.name}>
       <svg xmlns="http://www.w3.org/2000/svg" viewBox="0, 0, 400,400">
         <circle cx="200" cy="200" r="200" fill="#2cb44c" />
         <path


src/components/icon/icon.svelte
```diff
@@ -4,13 +4,23 @@
 </script>
 
 <script lang="ts">
-  export let name: string = undefined
+  import type { IconNames, StringWithAutocomplete } from '../types'
+
+  export let name: StringWithAutocomplete<IconNames> = undefined
   export let forceColor: boolean = false
+  export let size: string = 'unset'
+  export let color: string | undefined = 'unset'
```
petemill, line 12: Can you support this with an example of where this is more better than defining size and color in css? Every other component / html element is styled via css (or inline style if you must!), with all the features css provides (cascade, media queries, hover and focus states). If this is a developer-efficiency thing, then I'm just struggling to think of scenarios where there isn't at least some CSS in an app, unless it's a prototype. Or solved with tailwind, because that would provide similar functionality to every element, e.g. `<p class="color-red hover:color-darkred">`. With regular css, it doesn't even need to be provided to just this element but can be provided by a parent, e.g.

```css
.toolbar {
  --leo-icon-size: 24px;
  color: blue;
}
```
or
```css
.toolbar {
  --leo-icon-size: 32px;
  @container (width < 321px) {
    --leo-icon-size: 24px;
  }
}
```
or
```css
.icon {
  color: pink;
  &:hover {
    color: blue;
  }
}
```
or
```css
.indicator {
   &.isError .icon {
      color: red;
   }
   &.isWarning .icon {
      color: yellow;
   }
}
```

src/components/icon/icon.svelte
```diff
@@ -39,10 +51,14 @@
     /* Non color icons are set via a mask-image, so we can change the color
      * by setting the background */
     & .icon:not(.color) {
-      background: currentColor;
+      background: var(--icon-light-mode-color);
       -webkit-mask-image: var(--icon-url);
       mask-image: var(--icon-url);
       mask-repeat: no-repeat;
+
+      @media (prefers-color-scheme: dark) {
```
petemill, line 59: This would need to be `@theme (dark)` to use our postcss plugin which accounts for dark mode "containers" provided by, e.g. `<div data-theme="dark">` wrappers for entire UI areas.

petemill (commented):
Perhaps we should split out the name type change and the more "controversial" (😄) style props?

src/components/icon/icon.svelte
AlanBreck: The CSS approach would also mean that we don't need a special `darkModeColor` prop since the consumer can change the color via its own media queries however it pleases.

josheleonard: This is how I am seeing Icon instance colors and sizes getting set from within React.

Screenshot 2023-03-23 at 2 05 17 PM

petemill: Seems like that should work. Is the issue that it's very verbose and that you have to make a new element for each time we want to change the color or size of something? If so, that's a problem I too have with styled-components in general - same is true every time we want a different font-size or color for a piece of text...
But it does look like some of the above can be simplified. For example `LinkIconCircle` can contain 
```
  --leo-icon-size: 16px;
  color: ${leo.color.white};
```
without having to make a whole new LinkIcon class.

Same for `BackButton` (and if you use Leo's `<Button>`, it will automatically set a good `--leo-icon-size` based on the button type.
And maybe something similar would apply to the component that `InfoIcon` is in?

src/components/icon/icon.svelte
```diff
@@ -4,13 +4,23 @@
 </script>
 
 <script lang="ts">
-  export let name: string = undefined
+  import type { IconNames, StringWithAutocomplete } from '../types'
+
+  export let name: StringWithAutocomplete<IconNames> = undefined
```
fallaciousreasoning, line 9: 👍 this is sweet!

src/components/types.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2023 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import type meta from '../../icons/meta'
+
+export type StringWithAutocomplete<T> = T | (string & Record<never, never>)
+
+export type IconNames = keyof (typeof meta)['icons']
```
fallaciousreasoning, line 10: 👍

petemill, line 10: How about we put this in `icons/meta.ts`? That's where the value is and its utility is not only for components. And then how about we clear up the name of this file, since it's not all types in components as they are usually defined with each component. How about `utlity.ts` or `utility_types.ts` or something in that nature?

icons/meta.ts
```diff
@@ -1778,4 +1778,4 @@ export default {
     'zilliqa-color': 'zilliqa-color',
```
petemill, line 1778: This file is auto-generated by `src/scripts/update-icons.js` so we'll need to make this adjustment there too.

josheleonard:
Closing since I do not currently have access to API keys needed to test `src/scripts/update-icons.js`

fallaciousreasoning:
@josheleonard if you have access to the Figma files you should be able to get one - (click `request personal access token` [here](https://www.figma.com/developers/api)) and then just set it with `export FIGMA_API_TOKEN=<your_token>` before running the script.

Alternatively, you can probably run this without doing the download from Figma:
1. Comment out lines 126 - 140 in `src/scripts/update-icons.js`
2. Add a call to `generateMeta()`
3. Run `node src/scripts/update-icons.js`

And the meta file should get regenerated.

```diff
diff --git a/src/scripts/update-icons.js b/src/scripts/update-icons.js
index 0a02e8b..23281d7 100644
--- a/src/scripts/update-icons.js
+++ b/src/scripts/update-icons.js
@@ -10,11 +10,11 @@ const FINAL_FOLDER = './icons'
 fs.mkdirSync(RAW_FOLDER, { recursive: true })
 fs.mkdirSync(FINAL_FOLDER, { recursive: true })
 
-if (!process.env.FIGMA_API_TOKEN) {
-  throw new Error(
-    `In order to work, this script requires a figma API token to be set in the 'FIGMA_API_TOKEN' environment variable. You can get one from here https://www.figma.com/developers/api#access-tokens and set it with 'export FIGMA_API_TOKEN=<YOUR_API_TOKEN>'`
-  )
-}
+// if (!process.env.FIGMA_API_TOKEN) {
+//   throw new Error(
+//     `In order to work, this script requires a figma API token to be set in the 'FIGMA_API_TOKEN' environment variable. You can get one from here https://www.figma.com/developers/api#access-tokens and set it with 'export FIGMA_API_TOKEN=<YOUR_API_TOKEN>'`
+//   )
+// }
 
 const getFlag = (iconName) => {
   const regionRegex = /Country=Region - ((\w|\s)+).svg/
@@ -123,17 +123,18 @@ const generateMeta = () => {
 
 const exporter = figmaApiExporter(process.env.FIGMA_API_TOKEN)
 
-exporter
-  .getSvgs({
-    fileId: 'g8Z0q6TMPYDq6zXh9Y7LWD',
-    canvas: '🔮 Icons'
-  })
-  .then(async (svgsData) => {
-    await exporter.downloadSvgs({
-      saveDirectory: RAW_FOLDER,
-      svgsData: svgsData.svgs,
-      lastModified: svgsData.lastModified
-    })
-  })
-  .then(mutateIcons)
-  .then(generateMeta)
+// exporter
+//   .getSvgs({
+//     fileId: 'g8Z0q6TMPYDq6zXh9Y7LWD',
+//     canvas: '🔮 Icons'
+//   })
+//   .then(async (svgsData) => {
+//     await exporter.downloadSvgs({
+//       saveDirectory: RAW_FOLDER,
+//       svgsData: svgsData.svgs,
+//       lastModified: svgsData.lastModified
+//     })
+//   })
+//   .then(mutateIcons)
+//   .then(generateMeta)
+generateMeta()
```

src/components/icon/icon.svelte
fallaciousreasoning: Thinking about this a bit more it could be just wrapped in a styled component with props?

```ts
export const LinkIcon = styled.div<{ size: string, color: string }>`
    --leo-icon-size: ${p => p.size};
    color: ${p => p.color};
```

but if we're doing this, maybe we just want to expose it on our button? I think if this is going to be good for developer experience, I'd be in favor of doing it. In addition, it'll make it much easier to tell what the icon is going to look like:

```tsx
<Icon color={leo.color.white} size={leo.icon.size[16] name="link"/>

// as opposed to

const WhatIDoNowIcon = styled(Icon)`
    --leo-icon-size: ${leo.icon.size[16};
    color: ${leo.color.white};
`

/* a whole bunch of code which isn't related to icons at all */

<WhatIDoNowIcon name="link"/>
```

**controversial opinion:**
I guess what I'm saying is that I don't see the harm in adding these style properties, if they're going to be super common

PR #247, opened 2023-03-23
updates renovate schedule config

AlanBreck:
It appears from [the Renovate docs](https://docs.renovatebot.com/configuration-options/#schedule) that the `schedule` config expects an array.


fallaciousreasoning (approved):
:facepalm: thanks @AlanBreck

PR #249, opened 2023-03-26
WebComponents Followup

fallaciousreasoning:
This PR fixes a few issues that @petemill & @AlanBreck brought up in the original [PR](https://github.com/brave/leo/pull/239) which weren't fixed because we were in a hurry to land it.

### DIFF
--- a/src/components/svelte-web.ts
+++ b/src/components/svelte-web.ts
@@ -5,8 +5,15 @@ interface Options {
   name: string
 }
 
+// Properties with these types should be reflected to attributes.
 const reflectToAttributes = new Set(['string', 'number', 'boolean'])
 
+/**
+ * This function creates a faux Svelte component which forwards WebComponent
+ * slots into a Svelte slot.
+ * @param name The name of the slot
+ * @returns A Svelte "component" representing the slot.
+ */
 const createSlot = (name?: string) => {
   let slot: HTMLElement
   return {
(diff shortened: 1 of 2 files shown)


src/components/svelte-web.ts
```diff
@@ -217,8 +238,8 @@ export default function registerWebComponent(
     }
 
     removeEventListener(event: string, callback: Callback) {
-      this.listeners.get(event)?.get(callback)?.()
-      this.listeners.get(event)?.delete(callback)
+      this.listenerRemovers.get(event)?.get(callback)?.()
+      this.listenerRemovers.get(event)?.delete(callback)
```
AlanBreck, line 242: Maybe a silly question, but should the `remover` function also be modified to always remove the callback from `listenerRemovers`?

AlanBreck:
This PR makes me wonder if we can also publish our examples somewhere, just like we publish Storybook. Maybe like `https://249.pr.leo.bravesoftware.com/examples`?

src/components/svelte-web.ts
fallaciousreasoning: We could, I don't think it's critical though. On line 85 we don't need it to remove itself from the listeners array (though doing so wouldn't hurt) because we overwrite it in `addEventListener`. I could go either way on this though, wdyt is best

petemill (approved):

PR #254, opened 2023-03-30
[Bug]: Fix mobile tap highlights

fallaciousreasoning:
Currently on mobile, tapping anything with `cursor: pointer` will result in a nasty blue rectangle on WebKit/Chromium browsers. The solution is to just set `-webkit-tap-highlight-color: transparent` when we set `cursor: pointer`.

Before:
image
image

After
image
image


AlanBreck (approved):
LGTM! That said, this seems like another case where it'd be nice to have a Leo normalizer that just handles this kind of thing so that we don't have to think about it in components development.

PR #256, opened 2023-04-02
[Icons]: Updated icons from @aguscruiz 

fallaciousreasoning:


petemill:
hmm, prettier doesn't seem to like icons/meta.js @fallaciousreasoning

fallaciousreasoning:
Forgot to run format - ideally we wouldn't need to but for now, this is fine.

petemill (approved):

PR #259, opened 2023-04-18
Add action for auto updating symbols

fallaciousreasoning:
@kim0 I've created an action in the Leo SF Symbols repo which I want to run on deploy - would you mind having a look at this and letting me know if you think it's the right approach (and if it is, setting up the `LEO_SF_SYMBOLS_ACCESS_TOKEN` as a token which will let the action get triggered).

I haven't done much with GH Actions before so I might be going in the completely wrong direction.

### DIFF
--- a/.github/workflows/sf-symbols.yml
+++ b/.github/workflows/sf-symbols.yml
@@ -1,0 +1,23 @@
+name: Update SF Symbols
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+
+    steps:
+    - uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.LEO_SF_SYMBOLS_ACCESS_TOKEN }}
+        script: |
+          await github.rest.actions.createWorkflowDispatch({
+            owner: 'brave',
+            repo: 'leo-sf-symbols',
+            workflow_id: 'update.yml',
+            ref: 'main'
+          })


.github/workflows/sf-symbols.yml
```diff
@@ -0,0 +1,23 @@
+name: Update SF Symbols
+on:
+  push:
+    branches:
+      - main
```
kim0, line 5:
```suggestion
  push:
    paths:
      - "icons/**"
    branches:
      - main
```

.github/workflows/sf-symbols.yml
```diff
@@ -0,0 +1,23 @@
...
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
```
kim0, line 11: Let's delete this since it's not needed

.github/workflows/sf-symbols.yml
```diff
@@ -0,0 +1,23 @@
...
+
+    steps:
+    - uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.LEO_SF_SYMBOLS_ACCESS_TOKEN }}
```
kim0, line 16: Don't really like tokens, but I don't think we can trigger another repo without it.
cc @mihaiplesa to check and maybe create the token as I don't have permission on that repo.

kim0 (approved):
LGTM. Left some comments

.github/workflows/sf-symbols.yml
kim0: Only referring to the `permissions` line

kim0:
@fallaciousreasoning CI seems to need a fix too

.github/workflows/sf-symbols.yml
fallaciousreasoning: Wow awesome! Thanks @kim0

mihaiplesa (approved):

fallaciousreasoning:
I'm going to merge - dunno if it's going to work :laughing:

fallaciousreasoning:
Wonderful, thanks @kim0!

Issue #71, opened 2022-11-02
Component: Modal window


aguscruiz:
https://www.figma.com/file/H11ZOl6JMYbCTW4ZJXqR5V/%F0%9F%96%A5-Desktop-%26-Web?node-id=138%3A10797

![image](https://user-images.githubusercontent.com/15734630/199500492-4a38f4a5-3789-43b6-85e5-a0908b54b24f.png)

OAuthBringer:
Required for Creators 2.0

fallaciousreasoning:
@aguscruiz I can't seem to find this component in Figma

aguscruiz:
Sorry! Must've moved it. Here it is:
https://www.figma.com/file/445ratdeFt26kKk53LSMj0/%F0%9F%92%BB-Websites?node-id=1-6363

fallaciousreasoning:
I'm working on this at the moment. Here are some notes from out last meeting:

## Dialog vs Modal
Dialogs and Modals are very similar, but modals have a backdrop while dialogs don't. We'll handle this with a `modal` property which can be added to a dialog.

In addition, it **might** be possible to scroll the content behind a dialog (but not a modal). Dialogs also (currently) only support the `small` size, while dialogs also have a bigger mode.

`back`/`close` button should be supported via boolean flags. The back button should fire a `back` event, and is intended for use in multi step wizards.

## Slots

The component should have the following slots:
- Title
- Subtitle (doesn't affect layout when not present)
- Body
- Actions

## Imperative API

It should be possible to use the component via an imperative API, allowing us to use the dialog component instead of the `window.alert`, `window.confirm` and `window.prompt` functions.

PR #260, opened 2023-04-19
[Component]: Dialog

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/71

@petemill what do you think of https://github.com/brave/leo/pull/260/files#diff-4a25d419101f95ef68e52670ca1834d529de65c0f2d2b216ddc3a0cab6136b15R6-R17, which automatically creates a container for Dialog popups, rather than you needing to add one to the dom?

If you think it's a good idea I can go back to the alert centre and use the same pattern there.

@aguscruiz there are a few areas where I think I need your input on defaults.
1. Should dialogs have a close button by default (currently no)
2. Should dialogs have a back button by default (currently no)
3. Should dialogs be closable with the escape key by default (currently yes)
4. Should dialogs be closable by clicking outside by default (currently yes)


src/components/dialog/dialog.svelte
```diff
@@ -0,0 +1,180 @@
...
+    background: var(--leo-color-container-background);
+    padding: var(--leo-spacing-32) var(--leo-spacing-32) var(--leo-spacing-40)
+      var(--leo-spacing-32);
+  }
+
+  .leo-dialog.mobile {
+    width: 374px;
+  }
+
+  .leo-dialog:not(.modal) {
```
petemill, line 118: Doesn't it always have the class `modal`?

src/components/dialog/dialog.svelte
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  :global .leo-dialog .actions ::slotted(*),
+  :global .leo-dialog div[slot='actions'] {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    justify-content: stretch;
+    gap: var(--leo-spacing-16);
+  }
```
petemill, line 170: Perhaps a comment here explaining what these `:global` commands are attempting to do?

src/components/dialog/dialog.svelte
```diff
@@ -0,0 +1,180 @@
...
+  on:click|capture={(e) => {
+    if (!dialog) return
+
+    const rect = dialog.getBoundingClientRect()
+    const clickedOutside =
+      e.clientX < rect.x ||
+      e.clientY < rect.y ||
+      e.clientX > rect.x + rect.width ||
+      e.clientY > rect.y + rect.height
+    if (isOpen && clickedOutside && backdropClickCloses) {
```
petemill, line 90: how about early return if these conditions aren't met? That way we can avoid the `getBoundingClientRect` call

src/components/dialog/dialogHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export const confirm = (message: string, options?: ConfirmDialogOptions) =>
+  dialog({
+    title: message,
+    actions: [
+      {
+        result: false,
+        kind: 'outline',
+        text: options?.cancelText ?? 'Cancel'
```
petemill, line 84: I believe(?) this is the first time we're introducing some default text in to a Leo component, so we should probably think about internationalisation. Perhaps a provider-based approach where consuming apps can provide the strings that all Leo components need, like:
```ts
type LeoStrings = { cancel: string, ok: string }

Leo.ProvideStrings = function(strings: LeoStrings)
```

Not sure if we can get away with sending the current locale only or send all locales and also tell leo which is the current locale...

src/components/dialog/dialogHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+    title: options.title,
+    body: options.body,
+    actions: options.actions,
+    resolve: (result?: T) => {
+      dialogs.update((d) => d.filter((i) => i !== info))
+      resolve(result)
+    }
+  }
+
+  dialogs.update((d) => [...d, info])
```
petemill, line 53: Technically this part could be done inside the Promise callback, could it not, avoiding both the `resolve` assignment and the possible but extremely unlikely possibility that resolve is called before it's assigned. Although I'm not sure when the Promise callback is called so it _might_ actually be impossible that happens...

petemill (approved):
Some interesting techniques - and sure why not offer both ways to show a dialog if it works well!

src/components/dialog/dialog.svelte
fallaciousreasoning: Wow, so it does read like that.

I'd originally written
`class:modal={modal}` but when I format the file it gets changed to `class:modal`, which I guess is shorthand (it still seems to work).

src/components/dialog/dialogHelpers.ts
fallaciousreasoning: I think the earliest a promise can resolve is in the next tick, so it shouldn't ever happen. That said, yours reads nicer :laughing:

src/components/dialog/dialogHelpers.ts
fallaciousreasoning: I agree, we should definitely think about this. For now though, I'm going to kick the can down the road and force you to provide strings.

Issue #261, opened 2023-04-19
How Customizable should controls be?


fallaciousreasoning:
It's common to have slightly different usecases for common controls, where the default styling is *almost* but not quite right. In these cases, it's useful for consumers to be able to customize parts of the control, so they get the benefits of the built in Leo control (including accessibility, keyboard control, default styling, theme support).

An example given recently was customizing the label on our Checkbox, where we want to be able to build the following layout:
![image](https://user-images.githubusercontent.com/7678024/233192536-597b2f78-70f2-4ea0-96d4-b34d894575f1.png)

This is not currently possible.

Possible approaches:
1. Expose a `--leo-checkbox-direction` which supports different flex directions?
2. Expose a css `::part`?
3. Expose a bunch of properties from the control?

@petemill @aguscruiz @AlanBreck

PR #266, opened 2023-04-27
Make it possible to change the direction of radios/checks

fallaciousreasoning:
This fixes (partially) https://github.com/brave/leo/issues/261, by making it possible to reverse the direction of the label/checkbox. However, we should definitely think about this some more.

This fixes the issue raised here: https://bravesoftware.slack.com/archives/C03A0MWJXP0/p1681934571717649

### DIFF
--- a/src/components/checkbox/checkbox.stories.svelte
+++ b/src/components/checkbox/checkbox.stories.svelte
@@ -18,6 +18,11 @@
       type: 'string',
       description: 'The gap between the label and the checkbox'
     },
+    '--leo-checkbox-flex-direction': {
+      control: 'select',
+      description: 'The flex direction of the label/checkbox',
+      options: ['row', 'row-reverse', 'column', 'column-reverse']
+    },
     '--leo-checkbox-checked-color': {
       control: 'color',
       type: 'string',
--- a/src/components/checkbox/checkbox.svelte
+++ b/src/components/checkbox/checkbox.svelte
@@ -64,6 +64,7 @@
   .leo-checkbox {
     --focus-border-radius: var(--leo-checkbox-focus-border-radius, 2px);
     --label-gap: var(--leo-checkbox-label-gap, var(--leo-spacing-8));
+    --flex-direction: var(--leo-checkbox-flex-direction, row);
     --checked-color: var(
       --leo-checkbox-checked-color,
       var(--leo-color-icon-interactive)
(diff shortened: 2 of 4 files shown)


src/components/checkbox/checkbox.stories.svelte
```diff
@@ -18,6 +18,11 @@
       type: 'string',
       description: 'The gap between the label and the checkbox'
     },
+    '--leo-checkbox-flex-direction': {
+      control: 'select',
+      description: 'The flex direction of the label/checkbox',
+      options: ['row', 'row-reverse', 'column', 'column-reverse']
```
AlanBreck, line 24: I lean towards changing this to something like `--leo-checkbox-direction` or `--leo-checkbox-label-position` and not allowing `column` since it would be a strange, likely never used, look.

fallaciousreasoning:
@petemill I had a chat with @AlanBreck and we weren't sure whether it's useful to be able to change the `flex-direction` without also being able to change the `align-items` and the `justify-content` properties.

It feels a bit odd exposing all of this, I wonder if we wouldn't be better just exposing a CSS `::part`? What do you think?

fallaciousreasoning:
Talked to @petemill and we reckon exposing `justify-content` and `flex-direction` is probably fine for now :smile:

AlanBreck:
> Talked to @petemill and we reckon exposing `justify-content` and `flex-direction` is probably fine for now 😄

I have misgivings, but am totally willing to acquiesce. :-) Can you give the nutshell of why?

fallaciousreasoning:
Yup, it definitely feels a bit odd, but basically, there are three ways we could go:
1. A prop for `orientation`
2. A CSS `::part`
3. Expose variables for `flex` stuff

@petemill reckons that the CSS `::part` will let people customize the control too much, and we might break it when we update. Everyone is a little bit wary of adding a bunch of props for customizing the control, so I guess that kind of leaves us with exposing the CSS variables

@petemill also had an idea that we could us container queries to automatically change the `justify-content` to `space-between` when it hits a certain width but I'm a little worried that we'll choose the wrong width. Basically, this is what we're left with :D

petemill:
I think you summarised it well @fallaciousreasoning. What are your misgivings @AlanBreck? Is it simply to also expose alignment? Isn't the alignment just between the control and the label, and not related to anything external? Won't we always want it the same? (...which I'm imagining is the middle of first line of label, but not middle of whole label...)

PR #269, opened 2023-04-28
[Icons]: Fix for reactivity bugs

fallaciousreasoning:
This fixes a few bugs in the icon:
1. The `iconBasePath` is changed after the `Icon` is mounted, which wouldn't trigger a rerender
2. The icon changes from `color` to `non-color`, or vice versa, and the icon wouldn't detect the change.
3. Centre non-square icons (like the flags) so they don't look off

### DIFF
--- a/src/components/icon/icon.svelte
+++ b/src/components/icon/icon.svelte
@@ -66,6 +66,7 @@
       -webkit-mask-image: var(--icon-url);
       mask-image: var(--icon-url);
       mask-repeat: no-repeat;
+      mask-position: center;
     }
 
     /* Color icons need to be set via a background rather than a mask so we
@@ -73,6 +74,7 @@
     & .icon.color {
       background: var(--icon-url);
       background-repeat: no-repeat;
+      background-position: center;
     }
   }
 </style>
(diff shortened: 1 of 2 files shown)


src/components/icon/icon.svelte
```diff
@@ -1,25 +1,34 @@
...
-  export const setIconBasePath = (basePath: string) => (iconBasePath = basePath)
+  import { writable } from 'svelte/store'
 
-  const getIconUrl = (name: string) => `${iconBasePath}/${name}.svg`
+  let lastIconBasePath = '/icons'
+  let iconBasePath = writable(lastIconBasePath)
+
+  export const setIconBasePath = (basePath: string) => {
+    lastIconBasePath = basePath
+    iconBasePath.set(basePath)
```
AlanBreck, line 9: There's no particular reason why you _should_ do this, but you could also set this by writing `$iconBasePath = basePath`.

AlanBreck (approved):

src/components/icon/icon.svelte
fallaciousreasoning: I get this error
image

when I try and do that - I think it's because I'm not actually in the component logic here, I'm in the `module` so Svelte's reactivity stuff doesn't work :/

AlanBreck: Oh! I guess that makes sense.

Issue #55, opened 2022-11-02
Component: Dropdowns / Selector


aguscruiz:
https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?node-id=278%3A12810

![image](https://user-images.githubusercontent.com/15734630/199498933-d1bebd97-e07f-4940-a414-65a888fe247d.png)

PR #270, opened 2023-04-28
[Component]: Dropdown

fallaciousreasoning:
Adds a dropdown component, and a `control` component, which will hopefull be useful for other inputs

Fixes: https://github.com/brave/leo/issues/55

This component ended up being kind of complicated to implement in a way that makes it easy to customize, as both a web component, and a Svelte component. What I've got here seems to work quite nicely though.

@aguscruiz some feedback on the design:
1. I think the hover states are too subtle, it's pretty hard to notice
2. I think we should have some additional styling for the currently selected option.


src/components/radioButton/radioButton.svelte
```diff
@@ -41,7 +41,7 @@
     }
 
     currentValue = value
-    dispatch('changed', { value })
+    dispatch('change', { value })
```
fallaciousreasoning, line 44: I noticed this event had the wrong name

tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
   "extends": "@tsconfig/svelte/tsconfig.json",
-  "include": ["src/components/**/*"],
+  "include": ["src/components/**/*", "src/directives/**/*"],
```
fallaciousreasoning, line 3: Needed so we can import typescript files from outside src/components

src/components/dropdown/dropdown.stories.svelte
```diff
@@ -0,0 +1,113 @@
...
+
+<Story name="Default" />
+
+<Story name="Left Icon" let:args>
+  <div class="container" style={getStyleFromArgs(args)}>
+    <Dropdown {...getNonStyleArgs(args)}>
+      <div slot="left-icon">
+        <Icon name="check-circle-outline" />
+      </div>
+      <leo-option>One</leo-option>
```
fallaciousreasoning, line 49: Ideally, I think we'd specify options with `option` elements but they don't support styling or any complex children. If I `clone` the content of the option, it won't be in the shadow DOM any more (and won't get it's styles from outside the shadow).

I could pass in an element like a `<div>` here, and it would work, but then you can't specify a `<div value="foo">` without triggering intellisense errors.

My solution has just been to use a made up element (which is treated like a div).

What do you think?

src/components/dropdown/dropdown.svelte
```diff
@@ -0,0 +1,244 @@
...
+    const option = options.find((option) => e.composedPath().includes(option))
+
+    // If the event was triggered for something which isn't an option don't fire
+    // a change event.
+    if (!option) return
+
+    // If the option element doesn't have a value, fallback to using the text
+    // content - this allows writing simplified options:
+    // i.e. <o>1</o>
+    value = option.getAttribute('value') ?? option.textContent
```
fallaciousreasoning, line 52: I wasn't sure if I should do this - maybe we don't want to fall back to the text content of the node? We could enforce that all children have the `value` attribute?

src/components/icon/icon.svelte
```diff
@@ -56,13 +56,15 @@
       -webkit-mask-image: var(--icon-url);
       mask-image: var(--icon-url);
       mask-repeat: no-repeat;
+      mask-position: center;
```
fallaciousreasoning, line 59: This fixes a bug with non-square icons

examples/plain-html/index.html
```diff
@@ -80,7 +104,7 @@
     })
     dialog.addEventListener('close', () => dialogButton.isDisabled = false)
   </script>
-  <div class="component-container">
+  <leo-option class="component-container">
```
petemill, line 107: this doesn't seem right...?

src/components/dropdown/dropdown.svelte
petemill: I suppose it seems fine, but most of the time I'd imagine we'd have translated text.

src/components/dropdown/dropdown.stories.svelte
petemill: I don't think we need to worry too much about supporting `option` in this component, since we will have a select-like component with Input (which will support `option`), and the whole point of using this Dropdown component vs the regular Select component I imagine is only for complex children and properties. What do you think?

examples/plain-html/index.html
fallaciousreasoning: Wow, it was not :laughing: - I installed an extension for automatically editing the closing tag. I guess it has some bugs :laughing: 

Interestingly, it didn't affect the display at all - I guess the HTML parser is **really** forgiving

src/components/dropdown/dropdown.stories.svelte
fallaciousreasoning: Sweet, I guess what I have is fine for now then :smile:

src/components/dropdown/dropdown.svelte
fallaciousreasoning: Yeah - maybe it just doesn't buy us anything? Pretty happy to just remove it, if you prefer?

aguscruiz:
Looking great! A couple comments:

image
This is tricky, but when there's no icon, the left padding should be 16px instead of 8px. That's to give the text a bit more room to breathe.

---

image
- Icons should be 20px in the text fields
- Color should be the default icon color
- The field size will probably change with the new icon size, but the field itself should be 44px tall.

----

image

This looks great aside from the comment I responded in figma about using the pressed state. I don't know how to set icons in the dropdown menu but as long as they match the comments I left above, it should be good to go

fallaciousreasoning:
I always forget the cursors @aguscruiz :laughing: Thanks for taking a look :smile:

src/components/control/control.svelte
```diff
@@ -0,0 +1,106 @@
+<script lang="ts" context="module">
+  export let sizes = ['small', 'normal'] as const
+  export type Size = (typeof sizes)[number]
+</script>
+
+<script lang="ts">
+  export let disabled = false
+  export let size: Size = 'normal'
+  export let required = false
```
petemill, line 9: I like the idea of this - common functionality for form-style components. But perhaps "Control" is too generic and similar to "Component". I wouldn't classify a Checkbox as not a control, for example. How about "FormItem" or "LabeledComponent" or similar?

src/components/dropdown/dropdown.svelte
```diff
@@ -0,0 +1,254 @@
...
+
+    &.open {
+      transform: rotate(-180deg);
+    }
+  }
+
+  .leo-dropdown .leo-dropdown-popup {
+    background: var(--leo-color-container-background);
+    box-shadow: var(--leo-effect-elevation-03);
+    position: absolute;
```
petemill, line 198: This won't break out of `overflow: hidden` containers, right? I'm not saying you need to fix in this PR, but it has caused problems in the past so perhaps we can look at it in a follow-up, maybe using https://developer.mozilla.org/en-US/docs/Web/API/Popover_API.

Another issue that this can cause is knowing whether to open down or up and not having the popup being cut off by the viewport.

src/directives/clickOutside.ts
```diff
@@ -0,0 +1,21 @@
+export default function clickOutside(
```
petemill, line 1: `directives`? I like that it's not "utils" or something like that, but perhaps we need to make it clear this is web-specific, or put inside the components directory?

petemill (approved):
Works great. I'll leave it to you how you want to address my comments - they're pretty minimal, and to make sure @aguscruiz 's feedback is addressed too.

src/directives/clickOutside.ts
fallaciousreasoning: Directives is what svelte calls these functions 
https://svelte.dev/tutorial/actions

Maybe I'll rename the folder to `svelte-directives`?

src/components/dropdown/dropdown.svelte
fallaciousreasoning: Indeed it will not. That's a rabbit hole I wanted to avoid at first :laughing: 

The Popover API sounds great (and actually seems to work with the #experimental-web-platform-features flag) but I worry it doesn't have broad enough support yet.

src/components/control/control.svelte
fallaciousreasoning: Renamed to 'FormItem'

PR #271, opened 2023-05-01
fixes incorrect schedule config for renovate

AlanBreck:


fallaciousreasoning (approved):
This is the gift which keeps on giving :laughing:

Issue #54, opened 2022-11-02
Component: Input field / Text area / Dropdowns


aguscruiz:
https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?node-id=195%3A7599

![image](https://user-images.githubusercontent.com/15734630/199498846-bca49e30-802f-48c1-bbcc-81a7b3a6b650.png)

AlanBreck:
@aguscruiz, what's the difference between dropdowns here, and those in https://github.com/brave-experiments/leo/issues/55?

PR #272, opened 2023-05-02
[Component]: Input

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/54
Closes #79

### DIFF
--- a/src/components/formItem/formItem.svelte
+++ b/src/components/formItem/formItem.svelte
@@ -22,7 +22,7 @@
         description: 'The padding of the control',
         type: 'string'
       },
-      '--leo-control-font': { 
+      '--leo-control-font': {
         description: 'The font used by the control',
         type: 'string'
       },
@@ -86,7 +86,10 @@
     --padding: var(--leo-control-padding, 9px);
     --font: var(--leo-control-font, var(--leo-font-primary-default-regular));
     --leo-icon-size: var(--leo-control-icon-size, 20px);
-    --leo-icon-color: var(--leo-control-icon-color, var(--leo-color-icon-default));
+    --leo-icon-color: var(
+      --leo-control-icon-color,
+      var(--leo-color-icon-default)
+    );
     --gap: var(--leo-control-label-gap, var(--leo-spacing-s));
     --direction: var(--leo-control-label-direction, column);
 
--- a/src/components/formItem/formItem.svelte
+++ b/src/components/formItem/formItem.svelte
@@ -1,9 +1,44 @@
 <script lang="ts" context="module">
+  import type { ArgType } from '@storybook/addons'
   export let sizes = ['small', 'normal', 'large'] as const
   export type Size = (typeof sizes)[number]
 
   export let modes = ['filled', 'outline'] as const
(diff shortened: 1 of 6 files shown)


aguscruiz:
Couple comments:

- How about we use the text cursor on hover/active? 
- When you switch the input to have an icon, it gets a bit taller, it shouldn't change size. The right size is the one with no icon.

src/components/input/input.stories.svelte
```diff
@@ -0,0 +1,75 @@
...
+<Story name="Default" />
+
+<Story name="Character Count" let:args>
+  <div style={getStyleFromArgs(args)}>
+    <Input {...getNonStyleArgs(args)} bind:value={characterCountValue}>
+      {args.label}
+      <div slot="extra">{characterCountValue.length}/100</div>
+      <div slot="right-icon">
+        <Icon name="send" />
+      </div>
```
AlanBreck, line 59:
```suggestion
      <Icon name="send" slot="right-icon" />
```

src/components/formItem/formItem.svelte
AlanBreck, line 1: Possible improvement: can you remove the empty divs when no icons are provided?

fallaciousreasoning: Okay, so I had a quick play around with this - I think it actually makes things more complicated, as I need to:
1. Add another class for the central div, so we can apply the padding
2. Set display: flex; flex-direction: row; justify-content: stretch; to get the content to fill the div

So I've left this as is

fallaciousreasoning:
@petemill mentioned we should look at why there are two variables for showing errors

fallaciousreasoning:
Okay @petemill I've dug back into why we have two properties for determining whether we show errors

`hasErrors` indicates whether the component has any custom errors (i.e. `your password must contain the letter Z`)
`showErrors` determines whether the component should display the errors

We also have a required property, and we wanted to be able to manage that automatically. These are the states that we need to be able to handle:
1. A required input is empty, but the form has not been submitted (errors should not be shown)
2. A required input has text, has other errors, but has not been submitted (errors should not be shown)
3. A required input is empty and the form has been submitted (errors should be shown)
4. A required input has text, has other errors and has been submitted (errors should be shown)

This is how I was imagining it being used:

```tsx

const [errors, setErrors] = React.useState<string[]>([])
const [submitted, setSubmitted] = React.useState(false)

return <Input required showErrors={submitted} hasErrors={errors.length}>
    <ul slot="errors>{errors.map(e => <li>{e}</li>}</ul>
</Input>
```

petemill:
@fallaciousreasoning it makes sense to me, that's how I've done it in the past too. I just thought there were 3 properties for some reason!

petemill (approved):
Looks great, I think it was the errors slot that confused me as a third "prop" but it makes sense to have it separated and styleable.

petemill:
@mattwilliams85 noticed lack of `placeholder` prop. Seems neccessary. We can do in this PR before merging, or a follow-up

petemill:
Should other input props be passed on to the `<input`, especially aria? What about number related properties (min, max, regex, etc)?

PR #273, opened 2023-05-03
[Icons]: Fix update-icons.yml action

fallaciousreasoning:
`grep` uses regex syntax not `glob` syntax :facepalm:, so this was never detecting any updates

### DIFF
--- a/.github/workflows/update-icons.yml
+++ b/.github/workflows/update-icons.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Check for changes
         id: check-icons
         run: |
-          echo modified=$(`git status --porcelain | grep -q *.svg` && echo true || echo false) >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/update-icons.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Check for changes
         id: check-icons
         run: |
-          echo modified=$(`git status --porcelain | grep -q *.svg` && echo true || echo false) >> $GITHUB_OUTPUT
+          echo modified=$(`git status --porcelain | grep -q .svg$` && echo true || echo false) >> $GITHUB_OUTPUT
```
kim0, line 41: @fallaciousreasoning you might wanna replace the `.` with `\.` since the existing one matches any character, not necessarily a dot I believe

kim0 (approved):

PR #275, opened 2023-05-03
[Icons Update Script]: Stage path correctly

fallaciousreasoning:
@kim0 sorry to hit you with so many of these - nearly at the bottom of the file so there can't be too many more things I've messed up :laughing: 

Turns out I'm absolutely hopeless at writing workflow files


kim0 (approved):
It usually helps to run the pipelines in PR stage. Happy to help with that if still needed, but yeah this is probably the last one 😄

fallaciousreasoning:
I might just take you up on that :laughing:

fallaciousreasoning:
Thanks for all your help @kim0 - it works!
https://github.com/brave/leo/pull/276

PR #284, opened 2023-05-17
Fix bug in Skiafy for paths starting with a relative command

fallaciousreasoning:
For example, this fixes the `pwa-install` Skia Icon, making it look like this:
image

Instead of
image

This should unblock https://github.com/brave/brave-core/pull/18339

I made this when I was trying to work out what was going on, which I found quite useful:
https://github.com/fallaciousreasoning/skia-icon-editor
https://skia-icon-editor.vercel.app/

### DIFF
--- a/src/scripts/icons/skiafy.js
+++ b/src/scripts/icons/skiafy.js
@@ -11,6 +11,10 @@
  * http://www.apache.org/licenses/LICENSE-2.0
  */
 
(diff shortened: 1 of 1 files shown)


petemill (approved):
This bug is a great find! Nice digging.

AlanBreck (approved):

PR #286, opened 2023-05-25
[Dropdown]: Fixes for use in React

fallaciousreasoning:
1. Define `leo-option` element for use in Dropdown. This makes it possible to set from React
2. Remove `aria-selected` attribute - the web component was not regenerating these every change, like Svelte does
3. Disable pointer-events on the button - there was a weird Chromium only bug (which I think is just in Nightly) where it was triggering a double pointer event. It worked (and still works) fine in Firefox.


petemill (approved):
Very cool that the definition makes its way through to the different layers we have.

Douglashdaniel (approved):
++

Issue #290, opened 2023-06-06
skia/colors.h should be compiled with brave-core


sangwoo108:
Currently, it doesn't compile. 

In the switch-case clause in the file, kColorFooBar should be scoped with the enum class name => case Color::kColorFooBar. I guess we can update the script for this.

PR #291, opened 2023-06-07
Prefix enum class name before each cases

sangwoo108:
`enum class`'s member should be scoped with the name of `enum class`

fix https://github.com/brave/leo/issues/290


fallaciousreasoning (approved):
LGTM!

PR #299, opened 2023-06-16
Support gradient color for Icons

sangwoo108:
For some icons, we need gradient background. 
image

But It seems we can't pass the gradient value via `color: ...` and `currentColor`. 
So try plumbing a new property

From storybook > Icon > Default > set gradient string.

image

### DIFF
--- a/src/components/icon/icon.svelte
+++ b/src/components/icon/icon.svelte
@@ -28,6 +28,7 @@
   import type { IconName } from '../../../icons/meta'
   export let name: IconName = undefined
   export let forceColor: boolean = false
+  export let gradient: string = undefined
   $: hasColor =
     name?.endsWith('-color') || name?.startsWith('country-') || forceColor
 </script>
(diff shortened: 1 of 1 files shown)


petemill:
It looks like this PR doesn't really do anything different than what's currently available - taking a parameter and setting is as css background. In that case, I think it should just work without this PR using the `--leo-icon-color` css property. That's because in the current CSS, the value for `background` is decided as:
```css
--icon-color: var(--leo-icon-color, currentColor);
```
...meaning that `--leo-icon-color` value, if present, is priority over `currentColor` value.

Looks like you should just be able to set the css variable to a gradient and it should work. The icon will use color (or currentColor ) as a fallback, but it will prioritize the `--leo-icon-color` value if set. So you would do something like:

.myGradientIcon {
  --leo-icon-color: linear-gradient(...);
}

<LeoIcon className={'myGradientIcon'} />
or the styled-components way
const MyPlaylistIcon = styled(LeoIcon)`
  --leo-icon-color: linear-gradient(...);
`

<MyPlaylistIcon ... />

PR #303, opened 2023-06-27
[Icon]: Fix Icon types

fallaciousreasoning:
This was being caused because the types for the icons didn't live inside the `./src` folder which all the other types live inside, so when we copied the types to a different folder the relative path was breaking.

My fix here was to simply stop copying the types to a new folder, as I didn't feel like it was adding anything to our build process. Possibly we'll want to revisit this if we decide to start outputting preprocessed Svelte components, but for now, I think this is fine.


.gitignore
fallaciousreasoning, line 1: Note: I sorted the lines in this file.

src/scripts/gen-svelte-types.js
fallaciousreasoning, line 1: This file was renamed to `gen-types` but Git hasn't picked that up, because most of the file was deleted.

petemill (approved):
seems ok!

PR #307, opened 2023-06-29
Fix JS tokens

fallaciousreasoning:
Currently, the JS tokens don't exactly match up with the CSS variables. For example, the `desktop` portion of the CSS variables is stripped but not in the tokens. This leads to some JS tokens which don't exist. For example:

`font.desktop.primary.heading.h3: var(--leo-desktop-primary-heading-h3)`, which does not exist. This should be `font.primary.heading.h3: var(--leo-primary-heading-h3)`.

Additionally, this adds a test, so we'll know if we break this again.

### DIFF
--- a/src/tokens/transformation/web/formatCssAsJs.js
+++ b/src/tokens/transformation/web/formatCssAsJs.js
@@ -81,7 +81,10 @@ module.exports = ({ dictionary, file }) => {
   // The `desktop` segment is ignored in our property names, so make sure we
   // remove it from the properties, or this won't match up with the actual CSS
   // variables.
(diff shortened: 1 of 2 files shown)


petemill:
I was afraid of this. Ideally both implementations would use the same rewriting

petemill (approved):

fallaciousreasoning:
Yeah, I know, there are little bits and pieces all over the place which modify things. It's a bit scary. 

I think part of the problem is that the style dictionary transforms are all over the place too, and they're all expected to handle different languages. We might want to have a bit of a rethink about how we're doing the transforms.

PR #313, opened 2023-07-05
[Figma]: Add Variables

fallaciousreasoning:
Closes https://github.com/brave/leo/pull/344 (this includes those updates, but adds the variables back in, and tweaks the parser)


petemill:
Remove some key segments:
- Themeable
- Semantic

Don't include the light and dark versions of semantic and themeable, devs will access specific color values in the same way designers can, via `primitive`

petemill:
....maybe change "dialogs and modals" to "dialogs"

sangwoo108:
> Don't include the light and dark versions of semantic and themeable,

Hi, I'm wondering if I'm getting things properly. Are `semantic` colors independent on light/dark theme and should we be able to access it with `primitive`? Do we need to implement c++ script to port primitives?

fallaciousreasoning:
@sangwoo108 essentially, the switching is handled automatically for the CSS variables, and we don't want to expose a specific light & dark mode version of those colors (in CSS) anymore.

Currently, for the `--leo-color-red-10` we expose a `--leo-color-red-10` (which is dark/light aware), `--leo-color-dark-red-10` & `--leo-color-light-red-10`. In the new system, to work around some shortcomings in Figma variables, @aguscruiz has introduced a low level set of `primitive` colors, which aren't theme aware.

For example

`--leo-color-red-10` might be `--leo-color-primitive-red-10` in light mode and `--leo-color-primitive-red-90` in dark mode (though we directly use the RGBA values, rather than pointing to the primitive variables).

This shouldn't really affect the C++ colors too much, but:
1. I've exposed the primitive colors in our `colors.h` file (and via our `GetColor` function, though the light/dark variant is the same).
2. You can still access the separate `light`/`dark` tokens (though maybe we want to restrict this @petemill?)


Migrating to the new variables system in Figma has been a pretty major undertaking by @aguscruiz but once this PR lands, this is a rough breakdown of the changes:

1. The tokens in `semantic` & `themable` combine to make up the existing `dark`/`light` mode aware tokens. These are not affected by the migration (i.e. `--leo-color-red-10` is still `--leo-color-red-10`, as we strip the `semantic` and `themable` portion of the token).
2. `dark`/`light` specific tokens have been removed in favor of the primitive tokens (i.e. `--leo-color-light-red-10` is gone, replaced by `--leo-color-primitive-red-10`. The equivalent dark token is **probably** `--leo-color-primitive-red-90` or something :laughing:)
3. To a lesser degree, this affects spacing & radius tokens, except the sizes have changed a bit (which scares me).

petemill:
> You can still access the separate light/dark tokens (though maybe we want to restrict this @petemill?)

Yeah I think we should restrict light / dark access, but leave access to them via primitives. The same way as CSS. But we can do that in a follow-up so we don't delay this. Maybe even in https://github.com/brave/leo/pull/343

fallaciousreasoning:
@petemill & @aguscruiz & @AlanBreck mind taking a look at this? It's a bit of a big one....

aguscruiz:
Everything looks great for me! There are a couple nitpicks on some of the components but it's super minor and not related to variables. I'll make tickets when I have some time.

But all the variables seem to be matching correctly. Incredible job @fallaciousreasoning!

petemill (approved):
🎆

Issue #128, opened 2022-12-08
[Meta] Diff token changes on PR


AlanBreck:
It would be really helpful to have a GH action automatically diff the design tokens file on a PR so that we could get a comprehensive list of:
- New tokens
- Renamed tokens
- Moved tokens (e.g. `color.white` => `color.extended.white`)
- Deleted tokens

Figma exports include a `styleId` which could be used for most of the above:
```js
"white": {
  "description": "",
  "type": "color",
  "value": "#ffffffff",
  "blendMode": "normal",
  "extensions": {
    "org.lukasoppermann.figmaDesignTokens": {
      "styleId": "S:b4ae88014d8866fb4a0a127eb34986f999ab7266,",
      "exportKey": "color"
    }
  }
},
```

However, it retains the same `styleId` when moved to a new group, so we'd have to determine a different means of learning if it moved.

PR #314, opened 2023-07-05
[Meta]: Provide diff of `variables.css`

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/128

This works via a couple of steps:
1. ~~Upload the `variables.css` file as part of the storybook. This means we can fetch the current tokens from https://leo.bravesoftware.com/css/variables.css~~ Done as part of #315 
2. Run a diff of that `variables.css` file and the current one
3. Upload it, as part of a details/summary, so we can see the diff on the PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,23 +98,19 @@ jobs:
           fi
           echo "::endgroup::"
 
-      - name: Fetch Tokens for Main
-        id: fetch-tokens
-        uses: gautemo/fetch-api-data-action@1.0.2
-        with:
-          url: 'https://leo.bravesoftware.com/css/variables.css'
-          file: 'tokens/css/variables.old.css'
-
       - name: Diff Tokens
         id: diff
         run: |
-          echo TOKENS_DIFF="$(diff -u tokens/css/variables.old.css tokens/css/variables.css)" >> $GITHUB_OUTPUT
+          curl https://leo.bravesoftware.com/css/variables.css > tokens/css/variables.old.css 
+          diff -u tokens/css/variables.old.css tokens/css/variables.css > tokens/css/variables.diff
 
       - name: Post GitHub comment
         uses: actions/github-script@v6
         if: ${{ github.event_name == 'pull_request' }}
         with:
           script: |
+            const fs = require('fs/promises')
+            const diff = await fs.readFile('./tokens/css/variables.diff', 'utf-8')
             await github.rest.issues.createComment({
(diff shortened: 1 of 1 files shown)


fallaciousreasoning:
@petemill mind taking a look at this?

fallaciousreasoning:
@kim0 @petemill PTAL

.github/workflows/build.yml
```diff
@@ -55,6 +55,14 @@ jobs:
         run: |
           npm run build
 
+      - name: Diff Tokens
```
kim0, line 58: Not sure why we're having to manually download that file and diff against it. Doesn't git already know what changed. For example, wouldn't something like `git diff ./tokens/css/variables.css` get the diff assuming that file changed ?

.github/workflows/build.yml
```diff
@@ -55,6 +55,14 @@ jobs:
...
+        id: diff-tokens
+        run: |
+          curl https://leo.bravesoftware.com/css/variables.css > ./tokens/css/variables.old.css
```
kim0, line 61:
```suggestion
          curl -fsSL -o ./tokens/css/variables.old.css https://leo.bravesoftware.com/css/variables.css
```

.github/workflows/build.yml
```diff
@@ -55,6 +55,14 @@ jobs:
...
+          # Note: Diff has exit code 1 for differences, and there are always
+          # some differences because the file contains a timestamp.
+          diff -u ./tokens/css/variables.old.css ./tokens/css/variables.css > ./tokens/css/variables.diff || echo "Diff Completed"
```
kim0, line 64:
```suggestion
          diff -u ./tokens/css/variables.old.css ./tokens/css/variables.css > ./tokens/css/variables.diff || :
```

kim0: If you just want to silence `diff`, I think `|| :` makes the intent clearer

.github/workflows/build.yml
fallaciousreasoning: This file is generated as part of the build process. Basically, we want to make it visible on the PR how the generated file is changing.

.github/workflows/build.yml
fallaciousreasoning: That is exactly what I wanted to do but my bash-fu is weak :laughing: Thanks @kim0

kim0 (approved):

AlanBreck (approved):
Clever!

PR #317, opened 2023-07-06
[Audit]: Fix bug where I was ignoring component tokens

fallaciousreasoning:
I added a node_modules ignore, and it was working when I was installing the module from a folder. Obviously it broke when I installed Leo via Github (and into node_modules) :facepalm:

### DIFF
--- a/src/scripts/audit-tokens.js
+++ b/src/scripts/audit-tokens.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+
 const fs = require('fs/promises')
 const path = require('path')
 const { walk } = require('./common')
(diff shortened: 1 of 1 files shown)


src/scripts/audit-tokens.js
```diff
@@ -7,6 +8,7 @@ const tokenRegex = /--leo-([a-zA-Z0-9]|-)+/gi
 const ROOT_FOLDER = path.join(__dirname, '..', '..')
 const CSS_VARIABLES = path.join(ROOT_FOLDER, 'tokens', 'css', 'variables.css')
 const COMPONENTS_FOLDER = path.join(ROOT_FOLDER, 'src', 'components')
+console.log(COMPONENTS_FOLDER)
```
petemill, line 11: Need this?

petemill (approved):

src/scripts/audit-tokens.js
fallaciousreasoning: ...no.... :laughing:

Issue #318, opened 2023-07-06
Button border radius


petemill:
Currently, when the button increases in height (because a consumer puts in too much text, or styles that text manually with a larger text size), the button appears _too_ rounded. @aguscruiz has asked for the roundedness to be constant at it's "regular" size, i.e. 22px for the 44px button, 18px for the 36px button, etc...

PR #319, opened 2023-07-06
Modified border radius all button variants

nullhook:
Resolves https://github.com/brave/leo/issues/318
All button variants now have a constant border radius that is half of the button's height.

### DIFF
--- a/src/components/button/button.svelte
+++ b/src/components/button/button.svelte
@@ -100,7 +100,6 @@
     transition: background 0.12s ease-in-out, var(--default-transition);
     box-shadow: none;
     border: solid var(--border-width, 0px) var(--border-color, transparent);
-    border-radius: var(--leo-radius-full);
     background: var(--bg);
     color: var(--color);
     text-decoration: none;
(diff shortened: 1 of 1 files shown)


petemill (changes requested):
Needs to at least look the same as current master:
image

fallaciousreasoning:
This might be related, but the focus outline for the jumbo hero button doesn't look quite right:
image

The focus outlines on the `plain` & `plain-faint` buttons aren't the same either, but this is present in main (so I don't think we need to fix it here).
`plain`
image
`plain-faint`
image

Thanks for working on this @nullhook!

src/components/button/button.svelte
```diff
@@ -149,30 +149,35 @@
...
     --padding-y: 12px;
     --padding-x: 30px;
+    border-radius: 22px;
   }
   .leoButton.isJumbo {
     --icon-size: 24px;
     font: var(--leo-font-components-button-jumbo);
     --padding-y: 18px;
     --padding-x: 26px;
+    border-radius: 24px;
```
fallaciousreasoning, line 180: Inspecting the button in devtools, the button is `60px` high, so should this be `30px`?

fallaciousreasoning (approved):
LGTM, pending the fix for the jumbo buttons (which I think the `border-radius` tweak will do)

src/components/button/button.svelte
fallaciousreasoning, line 1: I think we're still setting the border-radius to var(--leo-radius-full) on line 103. It can probably be removed.

fallaciousreasoning (approved):
LGTM, and the Storybook looks good!

Let's land it
image

petemill (approved):

Issue #51, opened 2022-11-02
Component: Floating action buttons


aguscruiz:
FAB background color should be overrideable.
https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?type=design&node-id=278-12159&mode=design&t=7KkASY16Er9t35S7-0

![image](https://user-images.githubusercontent.com/15734630/199498515-7078bb0f-c9b1-4daf-8786-b6e1b05172f6.png)

fallaciousreasoning:
Should the background color still be overridable?

PR #322, opened 2023-07-10
[Component]: Floating action buttons

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/51

In addition:
- Adds new story for fab buttons
- Fixes the focus outline for the `plain-faint` variant (previously it was heavily rounded)


petemill (approved):
FAB

Issue #306, opened 2023-06-28
Add opacity to colors


petemill:
@aguscruiz has stated that the transparency is intentional, so we need our values to reflect that.

CSS and Skia both need the alpha values

AlanBreck:
Is the transparency supposed to be configurable? If so, the alpha could be a a css variable value on the token. This is what Tailwind does for their colors, and how it's configured for the tailwind tokens.

petemill:
For these ones I don't think that is is meant to be configurable - it's part of the token.

aguscruiz:
Yeah, not configurable in general. Just part of the token

fallaciousreasoning:
For what it's worth, we can make them configurable pretty easily, if we expose a separate variable for the opacity. Is that worth doing, or is it just going to bloat the CSS file?

aguscruiz:
I'm not sure, designers usually change a layer's opacity instead of the color when designing. The only cases are here in the variables. But if it's not too much problem, then why not, sure

petemill:
Closing because
1) Designers can't change opacity when using variables
2) we already bring in the alpha (via `rgba()`) for variables which have alpha.

boocmp:
In tokens/css/variables
![image](https://github.com/brave/leo/assets/73575789/a25434e6-aa19-422a-8e2d-c832fdda4267)

In tokens/skia/color.h
![image](https://github.com/brave/leo/assets/73575789/e85b7803-8d2a-4055-b677-e21f3d7a8f0a)

There is no alpha in colors.h

fallaciousreasoning:
Good catch! We're pretty terrible at keeping these in sync. Reopening for Skia

PR #329, opened 2023-07-12
Optionally add alpha channel to Skia colors

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/306

This will output `SkColorSetRGB` for colors with no opacity and `SkColorSetARGB` for colors with an alpha channel.


petemill (approved):

fallaciousreasoning:
Before merging @fallaciousreasoning you should check it compiles in Brave-Core

PR #331, opened 2023-07-14

kim0:

Ignoring icon default-false.svg because:
- has a raster image element
Ignoring icon glow-color.svg because:
- has a raster image element
- has a raster image element
- has a raster image element
- has a raster image element


icons/chatgpt-color.svg
fallaciousreasoning, line 1: @aguscruiz does this icon need to be `-color`? It looks like we should be able to change it :smile:

aguscruiz: Black is their brand color 🫥.

But that's a good point, I should probably be able to change the color of this one and also others like Twitter, Facebook, etc. since we could recolor them in our UIs. 

I'll update them to remove the -color.

aguscruiz: Done!

fallaciousreasoning: >Black is their brand color 🫥.

:fearful:

fallaciousreasoning (approved):
LGTM

PR #336, opened 2023-07-20
[Collapse]: Add CSS variables for tweaking collapse padding

fallaciousreasoning:
Adds two new variables for controlling the padding of the collapse summary and content.

### DIFF
--- a/src/components/collapse/collapse.svelte
+++ b/src/components/collapse/collapse.svelte
@@ -47,6 +47,15 @@
   }
 
   .leoCollapse {
+    --summary-padding: var(
+      --leo-collapse-summary-padding,
+      var(--leo-spacing-16) 18px
+    );
+    --content-padding: var(
+      --leo-collapse-content-padding,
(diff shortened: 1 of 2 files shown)


src/components/collapse/collapse.svelte
```diff
@@ -47,6 +47,15 @@
   }
 
   .leoCollapse {
+    --summary-padding: var(
+      --leo-collapse-summary-padding,
```
petemill, line 51: We call it a title in the props, but a summary here. Should they be the same?

fallaciousreasoning: I'm a little conflicted about this - I think it should be `summary`, because conceptually we could have more than just the title in here (i.e. a little subtitle). That said, we don't support slotting content in here (maybe we should though?).

In favor of keeping this as summary:
- We optionally have an icon in here (which is not part of the title)
- We display the open/close indicator here (which is not part of the title).

petemill (approved):

PR #337, opened 2023-07-24
[Icons]: Update type definitions for icons

fallaciousreasoning:
The current icons meta exports the type as the default export. This means you can't use the icon names, just the type (which doesn't exist at runtime).


petemill (approved):
nice

PR #341, opened 2023-07-25
Add #pragma once for header files

### LINKED ISSUE #340: Skia output header should have header guard
color.h seems to be missing header guard to avoid including multiple times

sangwoo108:
Fix #340

Usually we use #ifndef template, but generated resources seems to be okay with #pragma once. Actually, the path could be different on the client source code, `pragma once` makes sense.


fallaciousreasoning (approved):
LGTM

PR #343, opened 2023-07-26
skia token name enum can extend brave chromium enum

petemill:
Modifies the skia enum so that it can be used in the brave/chromium color mixer.

```diff
> #include "brave/browser/ui/color/brave_color_id.h"
> #include "chrome/browser/ui/color/chrome_color_id.h"
285c287,288
< enum class Color {
---
> enum class Color : ui::ColorId {
>   kLeoColorsStart = kBraveColorsEnd,
416c419,420
<   kColorLegacyDivider1
---
>   kColorLegacyDivider1,
>   kLeoColorsEnd
```

### DIFF
--- a/src/tokens/transformation/skia/templates/colors.h.template
+++ b/src/tokens/transformation/skia/templates/colors.h.template
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include "brave/browser/ui/color/brave_color_id.h"
+#include "chrome/browser/ui/color/chrome_color_id.h"
 #include "third_party/skia/include/core/SkColor.h"
 
 namespace leo::light {
@@ -26,8 +28,10 @@ enum class Theme {
   kDark
 };
 
-enum class Color {
-<%= groupedTokens.light.allTokens.map(prop => `  ${prop.name}`).join(',\n') %>
+enum class Color : ui::ColorId {
+  kLeoColorsStart = kBraveColorsEnd,
+<%= groupedTokens.light.allTokens.map(prop => `  ${prop.name}`).join(',\n') %>,
+  kLeoColorsEnd
 };
 
 constexpr SkColor GetColor(Color color, Theme theme) {


fallaciousreasoning (commented):
This LGTM but lets rope in a C++ expert.

src/tokens/transformation/skia/templates/colors.h.template
```diff
@@ -26,8 +28,10 @@ enum class Theme {
   kDark
 };
 
-enum class Color {
-<%= groupedTokens.light.allTokens.map(prop => `  ${prop.name}`).join(',\n') %>
+enum class Color : ui::ColorId {
+  kLeoColorsStart = kBraveColorsEnd,
+<%= groupedTokens.light.allTokens.map(prop => `  ${prop.name}`).join(',\n') %>,
+  kLeoColorsEnd
 };
```
sangwoo108, line 35: How about wrapping these E_CPONLY() and define macro so that we can add this to brave_color_ids?

Roughly,

```suggestion
#define LEO_COLOR_IDS \
<%= groupedTokens.light.allTokens.map(prop => `  E_CPONLY(${prop.name}) \
`).join(',\n') %>,
};
```

and in brave_colod_id

```diff
#define BRAVE_COLOR_IDS               \
    BRAVE_COMMON_COLOR_IDS            \
    BRAVE_SEARCH_CONVERSION_COLOR_IDS \
    BRAVE_SIDEBAR_COLOR_IDS           \
    BRAVE_SPEEDREADER_COLOR_IDS       \
    BRAVE_VPN_COLOR_IDS               \
    BRAVE_VERTICAL_TAB_COLOR_IDS      \
-    BRAVE_PLAYLIST_COLOR_IDS
+    BRAVE_PLAYLIST_COLOR_IDS          \
+    LEO_COLOR_IDS
};

```

sangwoo108: I wish we have leo_color_mixer generated too 😄

fallaciousreasoning: I think @petemill has a solution for that!

petemill: @sangwoo108 That was the first strategy I tried for this PR, but I got some errors. And ultimately it seemed ok for it to be a separate enum set, similar to how chromium does it with separate enum sets. I'll try again to see what the error is. In the meantime, here's the LeoColorMixer PR https://github.com/brave/brave-core/pull/19440/files

PR #357, opened 2023-08-22
adds explicit :visited style to button component to avoid conflicts with anchor styles

AlanBreck:
This adds explicit styling for the `:visited` state in order to avoid having styles from `link` apply when the `button` component is an `<a>` tag instead of a `<button>`.

Screenshot 2023-08-22 at 5 53 54 PM


fallaciousreasoning (approved):
LGTM

PR #364, opened 2023-08-28
[Button]: Make button configurable

fallaciousreasoning:


fallaciousreasoning:
Closing - we're not going to make the button this configurable - instead we'll add a MenuButton variant

PR #369, opened 2023-09-01
Avoid error during ssr compile

petemill:
This doesn't make SSR work, but it attempts to at least include these components in an SSR environment (i.e. nodejs) and not throw an error.

### DIFF
--- a/src/components/svelte-web.ts
+++ b/src/components/svelte-web.ts
@@ -46,6 +46,10 @@ export default function registerWebComponent(
   component: any,
   { name, mode }: Options
 ) {
+  if (!globalThis.customElements) {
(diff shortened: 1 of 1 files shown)


src/components/svelte-web.ts
```diff
@@ -46,6 +46,10 @@ export default function registerWebComponent(
   component: any,
   { name, mode }: Options
 ) {
+  if (!customElements) {
```
fallaciousreasoning, line 49: I think this will still error - we probably need to check globalThis.customElements

fallaciousreasoning (approved):
Lgtm in principle

fallaciousreasoning (approved):
LGTM

PR #370, opened 2023-09-01
fix: prevent window methods being called when SSR

mattwilliams85:
Allows Leo react components to work with an SSR application


fallaciousreasoning (approved):
LGTM

Issue #368, opened 2023-08-31
Menu component


petemill:
Let's take the menu-ness from Dropdown and offer it as a standalone component. @fallaciousreasoning 's idea is that the anchor would be provided via a slot.

PR #383, opened 2023-09-12
Added menu component

nullhook:
Resolves: https://github.com/brave/leo/issues/368

TODO:
1.~Accept anchor element as a slot~
2. ~Make menu accessible~


src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,63 @@
...
+  }}
+/>
+
+<Template let:args>
+  <div class="container">
+    <Menu value={undefined} {...args}>
+      <div slot="anchor">
+        <Icon name="more-horizontal" />
+      </div>
+      <div slot="label">{args.label}</div>
```
fallaciousreasoning, line 47: I don't think this component needs a label, or if it does, it should be part of the anchor

nullhook: makes sense. I've removed this for now. However, we might need this in the future. ex:

image

petemill: Do you mean the "Available language models" part? Would that just be a non-selectable item in the menu, provided in the default slot?

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,235 @@
...
+<div class="leo-menu">
+  <button
+    id="anchor-button"
+    class="click-target"
+    aria-haspopup="true"
+    aria-controls="menu"
+    tabindex="0"
+    bind:this={button}
+    on:click|stopPropagation={handleAnchorClick}
+  >
```
petemill, line 115: I guess we just need to make sure that the consumer doesn't provide their own button and only provides non-buttony content

src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,74 @@
...
+      defaultValue: 'Label'
+    },
+    size: {
+      control: 'select',
+      options: sizes
+    },
+    mode: {
+      control: 'select',
+      options: modes
+    }
```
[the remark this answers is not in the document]
nullhook: nope, removed

src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,74 @@
...
+    'gb': 'United Kingdom',
+    'ca': 'Canada'
+  }
+</script>
+
+<Meta
+  title="Menu"
+  component={Menu}
+  argTypes={{
+    ...cssProperties,
```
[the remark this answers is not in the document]
nullhook: removed

src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,74 @@
...
+      <leo-menu-item>
+        <div>Regenerate</div>
+      </leo-menu-item>
+      <leo-menu-item> Copy </leo-menu-item>
+      <leo-menu-item> Share </leo-menu-item>
+      <leo-menu-item>
+        <div class="item">
+          <div>New Chat</div>
+          <Icon name="plus-add" />
+        </div>
```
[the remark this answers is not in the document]
nullhook: fixed

nullhook:
@petemill added a "slots" story

Yes, that was the initial plan however the markup and quite different:

> The menu element's markup is different from that of the select's dropdown menu. The select menu has items with an accessibility state of 'selected' or 'not selected,' and the entire menu has a value. The standalone menu component created in this PR is a list of actionable items.


I think we can accept `leo-menu-item` if it's a standalone menu or `leo-option` if it's a select menu but then these needs to be differentiated roles for ex, `aria-selected` wont apply to menu items but leo-options, etc.

src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,74 @@
...
+  import SlotInfo from '../../storyHelpers/SlotInfo.svelte'
+  import Slot from '../../storyHelpers/Slot.svelte'
+  import { modes } from '../formItem/formItem.svelte'
+
+  const countries = {
+    'nz': 'New Zealand',
+    'au': 'Australia',
+    'gb': 'United Kingdom',
+    'ca': 'Canada'
+  }
```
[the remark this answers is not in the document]
nullhook: yes, removed

src/components/menu/menu.svelte
fallaciousreasoning: I think even if they provide their own button this will be fine, as it has `all: unset`, so the styling should be fine :) Might be a bit weird from an accessibility perspective though

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,235 @@
...
+    {#if slots.anchor}
+      <slot name="anchor" />
+    {:else}
+      <div>▶ Actions</div>
```
fallaciousreasoning, line 119: So for slots, you can just provide the default value inside the slot:

```svelte
<slot name="anchor">
    <div>▶ Actions</div>
</slot>
```

I don't know if this default makes sense though - actions seems quite specific, and it won't be translated or anything.

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,235 @@
...
+
+  $: menuItems = Array.from(
+    (
+      popup?.querySelector('.leo-menu-popup slot') as HTMLSlotElement
+    )?.assignedElements() ??
+      popup?.querySelectorAll('.leo-menu-popup > *') ??
+      []
+  )
+
+  $: {
```
fallaciousreasoning, line 36: It worries me a bit that we're duplicating all this logic - the code is a bit finicky because it has to consider a bunch of different things, and I can pretty easily see a scenario where we fix a bug in one of these but not the other. It'd be nice if we could factor some of this out

fallaciousreasoning (commented):
This looks pretty good, but I do worry about the amount of duplication between this and the Dropdown.

There's quite a bit of fiddly code for wrangling things across Svelte/Web Components/React and I worry about things getting out of sync between them.

I agree that semantically they're different, but from a code (and visual) standpoint, things seem quite similar. I wonder if it's worth trying to come up with a shared model?

@petemill wdyt? I'm also open to merging this in it's current form and trying to solve this in a separate PR, as @nullhook needs this

fallaciousreasoning:
If we do decide to treat them as separate models, maybe we can do something more traditional, than the `<leo-menu-item>` components (i.e, with separate <Menu> and <MenuItem> components), so we don't need to do all the hackery with querySelector?

petemill:
An example of one fork of the popup code getting enhancements and leaving the other fork in the dust is https://github.com/brave/leo/pull/384

Is it only the `aria-selected` that differs between the two? If so, that seems like 2 lines of code that could be controlled via a prop, e.g. `selectable`:

```svelte
  // When the options change, we should assign each one a tabindex - this let's
  // us handle changing the selection with the keyboard (tab or arrow keys).
  $: {
    for (const [option, index] of options.map((o, i) => [o, i] as const)) {
      option.setAttribute('tabindex', (index + 1).toString())
      if (selectable) {
        if (value === getValue(option)) option.setAttribute('aria-selected', '')
        else option.removeAttribute('aria-selected')
      } else {
         option.setAttribute('role', 'menu-item')
      }
    }
  }
```

nullhook:
No, not just aria-selected but others too

**Standalone menu**

the anchor button:
```
aria-haspopup="true"
aria-controls="menu"
```

the menu:
```
role="menu"
```

the menu item:
```
role="menuitem"
```

**Select menu**
the label:
```

```
the option:

src/components/menu/menu.svelte
nullhook: makes sense, removed

nullhook:
I have rebased this PR on https://github.com/brave/leo/pull/384

After the rebase and update, the Menu component is designed to work with 'leo-menu-item' or 'leo-option'. It includes aria-related properties that are added to the respective elements. Additionally, it handles the current selected value, which will be useful for select components.

After #384 merges, I'll modify Dropdown component to use Menu internally.

src/components/menu/menu.svelte
nullhook: We won't be duplicating it anymore. I'll use the Menu component after you merge the floating-ui PR.

src/components/button-menu/button-menu.stories.svelte
```diff
@@ -0,0 +1,63 @@
...
+  import Icon from '../icon/icon.svelte'
+  import SlotInfo from '../../storyHelpers/SlotInfo.svelte'
+  import Slot from '../../storyHelpers/Slot.svelte'
+</script>
+
+<Meta title="ButtonMenu" component={ButtonMenu} />
+
+<Template let:args>
+  <div class="container">
+    <ButtonMenu>
```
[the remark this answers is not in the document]
fallaciousreasoning: Might be worth doing a bit more setup so it toggles on/off in the story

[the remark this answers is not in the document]
fallaciousreasoning: Actually, it just looks like the build job hasn't been triggered - maybe it only runs for pulls against `main` which is a bit of a pain. Feel free to disregard.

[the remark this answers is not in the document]
fallaciousreasoning: Okay, when I checked out locally it was similar, and toggling the `isOpen` prop on the storybook didn't seem to do anything

src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,44 @@
...
+  import Icon from '../icon/icon.svelte'
+
+  let button
+</script>
+
+<Meta title="Menu" component={Menu} />
+
+<Template let:args>
+  <div class="container">
+    <button bind:this={button}> View page source </button>
```
fallaciousreasoning, line 14: I think let's just remove this button - not really needed for the`Menu` component

src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,44 @@
+<script lang="ts">
+  import { Meta, Story, Template } from '@storybook/addon-svelte-csf'
+
+  import Menu from './menu.svelte'
+  import Icon from '../icon/icon.svelte'
+
+  let button
```
fallaciousreasoning, line 7: Nit: remove

src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,44 @@
...
+</Template>
+
+<Story name="Default" />
+
+<style>
+  .container {
+    width: 250px;
+  }
+
+  .item {
```
fallaciousreasoning, line 38: Optional:
```suggestion
  leo-menu-item {
```

src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,44 @@
+<script lang="ts">
+  import { Meta, Story, Template } from '@storybook/addon-svelte-csf'
+
+  import Menu from './menu.svelte'
+  import Icon from '../icon/icon.svelte'
+
+  let button
+</script>
+
+<Meta title="Menu" component={Menu} />
```
fallaciousreasoning, line 10: Add `argTypes` prop and set the defaultValue for `isOpen` to `true` so the menu is open by default.

fallaciousreasoning: See the `label` prop in `input.stories.svelte`

src/components/menu/menu.stories.svelte
```diff
@@ -0,0 +1,44 @@
...
+        <div class="item">
+          <div>New Chat</div>
+          <Icon name="plus-add" />
+        </div>
+      </leo-menu-item>
+    </Menu>
+  </div>
+</Template>
+
+<Story name="Default" />
```
fallaciousreasoning, line 31: Probably worth adding a `slots` story, for completeness

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,234 @@
...
+  }
+</script>
+
+<script lang="ts">
+  import { createEventDispatcher } from 'svelte'
+  import clickOutside from '../../svelteDirectives/clickOutside'
+  import Floating from '../floating/floating.svelte'
+
+  export let isOpen = false
+  export let target: HTMLElement | undefined = undefined
```
fallaciousreasoning, line 23: Optional: I'm not 100% sure if you need the `= undefined` - you might be able to write:

```suggestion
  export let target?: HTMLElement
```

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,234 @@
...
+  export let currentValue: string | undefined = undefined
+
+  let dispatch = createEventDispatcher()
```
fallaciousreasoning, line 26: Nit:
```suggestion
  const dispatch = createEventDispatcher()
```

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,234 @@
...
+    )?.assignedElements() ??
+      popup?.querySelectorAll('.leo-menu-popup > *') ??
+      []
+  )
+
+  $: {
+    for (const [menuItem, index] of menuItems.map((o, i) => [o, i] as const)) {
+      menuItem.setAttribute('tabindex', (index + 1).toString())
+
+      if (menuItem.tagName === 'LEO-OPTION') {
```
fallaciousreasoning, line 49: This probably needs to be `LEO-MENU-ITEM` now

fallaciousreasoning: Optionally this could be a prop, or we could come up with some cleverer way of doing this

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,234 @@
...
+        } else menuItem.removeAttribute('aria-selected')
+      } else {
+        menuItem.setAttribute('role', 'menuitem')
+      }
+    }
+  }
+
+  // Note: we check isOpen !== undefined here so this is recalculated every time the
+  // dropdown is opened/closed.
+  $: minWidth =
```
fallaciousreasoning, line 61: :+1:

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,234 @@
...
+
+  function selectMenuItem(e: Event) {
+    // Find the option which was clicked on, if any.
+    const item = menuItems.find((item) => e.composedPath().includes(item))
+
+    // If the event was triggered for something which isn't an option don't fire
+    // a change event.
+    if (!item) return
+
+    if (item.tagName === 'LEO-OPTION') {
```
fallaciousreasoning, line 72: `leo-option` again

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,234 @@
...
+      currentValue = getValue(item)
+    }
+
+    isOpen = false
+
+    dispatch('select-item', {
```
fallaciousreasoning, line 78: :+1:

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,234 @@
...
+
+  function handleBlur(e) {
+    isOpen = false
+    dispatch('close')
+  }
+</script>
+
+<div class="leo-menu">
+  {#if isOpen}
+    <Floating {target} placement="bottom-start">
```
fallaciousreasoning, line 130: Nit: add `autoUpdate` prop - this means if the user scrolls around the menu will do it's best to stay in view, while remaining anchored

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,234 @@
...
+      </div>
+    </Floating>
+  {/if}
+</div>
+
+<svelte:window on:keydown={changeSelection} />
+
+<style lang="scss">
+  :host {
+    display: inline-block;
```
fallaciousreasoning, line 155: :+1: I always forget this

fallaciousreasoning (approved):
This looks great @nullhook - I've left a bunch of comments, but none of them are super critical.

Before landing, could you do the following:
1. Make sure the Dropdown component in the `plain-html` example works (and ideally the react one too)
2. Make sure your new ButtonMenu component works in the `plain-html` example (and ideally the react one too)

You don't need to commit changes to the example projects, but you're welcome to push them if you want. It's more of a quick sanity check.

src/components/menu/menu.svelte
nullhook: done

src/components/menu/menu.stories.svelte
nullhook: done

src/components/menu/menu.stories.svelte
nullhook: done

src/components/menu/menu.stories.svelte
nullhook: removed

src/components/menu/menu.svelte
nullhook: i don't see this in the current codebase and I get a syntax error

src/components/menu/menu.stories.svelte
nullhook: added

src/components/button-menu/button-menu.stories.svelte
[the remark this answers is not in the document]
nullhook: ButtonMenu doesn't have isOpen exposed though?

src/components/menu/menu.svelte
nullhook: Why LEO-MENU-ITEM? aria-selected is set for LEO-OPTION only, right?

fallaciousreasoning: Okay, sorry, I misunderstood how this was working - I assumed it was a holdover from the dropdown. Maybe instead of handling this via the type of the children, we could add a role prop to the `Menu` component, and do it based on that.

I worry that doing it based on the type of the children might be a bit non-obvious. WDYT @petemill?

src/components/button-menu/button-menu.stories.svelte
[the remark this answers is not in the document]
petemill: Maybe we're talking about `Menu` here and not `ButtonMenu`. That has `isOpen` exposed on Storybook but it doesn't do anything. I think we don't need `Menu` on Storybook as it isn't a design system component in itself. It seems fine to have `ButtonMenu` and `Dropdown` exposed on Storybook and not confuse anyone with `Menu`.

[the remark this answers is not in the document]
nullhook: That makes sense and I agree with Menu examples on storybook it basically ended up duplicating the logic of ButtonMenu. Anyways, I have removed Menu's storybook.

src/components/button-menu/button-menu.stories.svelte
```diff
@@ -0,0 +1,63 @@
...
+        </div>
+      </leo-menu-item>
+    </ButtonMenu>
+  </div>
+</Template>
+
+<Story name="Default" />
+
+<Story name="Slots" let:args>
+  <SlotInfo description="The dropdown provides several slots for customization">
```
petemill, line 30: "dropdown" reference

src/components/menu/menu.svelte
petemill: It would be simpler for the users to simply only have leo-menu-item, with a role prop for the container to check for these conditional aria attrubutes.

One minor consideration would be menus whose popups behave more like lists, e.g. a "..." menu for choosing an AI model. It's basically a dropdown without the input. Should we allow that mechanism via choosing <leo-option> vs <leo-menu-item>? Or should it be via the container? What about a menu that has a mix of selectable options and menu-items? Will our devs be aware of this enough?

Happy to merge it like this if you think it's the way to go, though IMO let's specify in the Storybook when to use which. Also, are we missing `role="option"` for `leo-option` items?

src/components/menu/menu.svelte
```diff
@@ -0,0 +1,234 @@
+<script context="module" lang="ts">
+  import type { HTMLAttributes } from 'svelte/elements'
+  declare global {
+    namespace JSX {
+      interface IntrinsicElements {
```
petemill, line 5: Should we move `leo-option` definiton here too?

petemill (approved):
We can merge to https://github.com/brave/leo/pull/384 as that's blocked on https://github.com/brave/reviews/issues/1400 but required for this branch

src/components/button-menu/button-menu.stories.svelte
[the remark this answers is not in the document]
petemill: I still see Menu in storybook 
image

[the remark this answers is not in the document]
petemill: Looks like we didn't get a CI build after your latest push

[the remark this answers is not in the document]
petemill: We can ignore that as it looks like it's due to https://github.com/brave/leo/pull/389

src/components/button-menu/button-menu.stories.svelte
nullhook: removed

src/components/menu/menu.svelte
nullhook: added

src/components/menu/menu.svelte
nullhook: I have added role=option for leo-option items. 

Let's see if the difference between 'leo-option' and 'leo-menu-item' is obvious to consumers, and I believe the first consumer for this would be AIChat. I have added a note regarding this in Storybook for both ButtonMenu and Dropdown.

PR #384, opened 2023-09-12
[Tooltip]: Add floating UI, a tooltip & convert the Dropdown to use the new floating component

fallaciousreasoning:
@petemill & @AlanBreck I think `floating-ui` is going to be super helpful going forward :smile:

### DIFF
--- a/examples/example-ui-react/src/App.tsx
+++ b/examples/example-ui-react/src/App.tsx
@@ -48,7 +48,13 @@ function App() {
             <leo-option value="bar">Bar</leo-option>
           </Dropdown>
         </div>
-        <Tooltip text="Hello World" visible={spinning} onVisibilitychange={() => { console.log("Closed")}}>
+        <Tooltip
+          text="Hello World"
+          visible={spinning}
+          onVisibilitychange={() => {
+            console.log('Closed')
+          }}
+        >
           <LeoButton href="#foo">Link button!</LeoButton>
         </Tooltip>
       </header>
--- a/examples/example-ui-react/src/App.tsx
+++ b/examples/example-ui-react/src/App.tsx
@@ -1,9 +1,16 @@
 import * as React from 'react'
 import '../../../tokens/css/variables.css'
 import LeoButton from '../../../react/button'
+import Tooltip from '../../../react/tooltip'
 import styles from './App.module.css'
 import Input from '../../../react/input'
 import Dropdown from '../../../react/dropdown'
+
+import { color } from '../../../tokens/css'
+
+color.white
+
+let f = '#FFFFFF'
 
 function App() {
   // Verify that we can change props and children (slots)
@@ -41,7 +48,9 @@ function App() {
             <leo-option value="bar">Bar</leo-option>
           </Dropdown>
         </div>
-        <LeoButton href="#foo">Link button!</LeoButton>
+        <Tooltip text="Hello World" visible={spinning} onVisibilitychange={() => { console.log("Closed")}}>
+          <LeoButton href="#foo">Link button!</LeoButton>
+        </Tooltip>
       </header>
     </div>
   )
--- a/examples/example-ui-react/src/App.tsx
+++ b/examples/example-ui-react/src/App.tsx
@@ -48,7 +48,13 @@ function App() {
           <div slot="anchor-content">ButtonMenu</div>
           <leo-menu-item>Llama2-13b</leo-menu-item>
           <leo-menu-item>Llama2-7b</leo-menu-item>
-          <div style={{ padding: '10px 0', fontSize: 12, borderTop: '1px solid blue' }}>
+          <div
+            style={{
+              padding: '10px 0',
+              fontSize: 12,
+              borderTop: '1px solid blue'
+            }}
+          >
             Coding
           </div>
           <leo-menu-item>Llama2-13b</leo-menu-item>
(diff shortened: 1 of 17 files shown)


examples/example-ui-react/src/App.tsx
```diff
@@ -1,10 +1,17 @@
...
 import styles from './App.module.css'
 import Input from '../../../react/input'
 import Dropdown from '../../../react/dropdown'
 
+import { color } from '../../../tokens/css'
+
+color.white
+
+let f = '#FFFFFF'
+
```
petemill, line 14: Is this for testing for errors? Although I'm a bit confused what `f` is used for 😄

package.json
```diff
@@ -6,6 +6,7 @@
   "type": "module",
   "dependencies": {
     "@ctrl/tinycolor": "3.5.1",
+    "@floating-ui/dom": "1.4.4",
```
petemill, line 9: let's get a security review for this

petemill (commented):
one thing I'm a little confused about is that the dropdown in storybook doesn't seem to be flipping to opening above the "input" if there isn't any room underneath in the viewport (or the document). Am I understanding flip and its default value correctly?

image

src/components/tooltip/tooltip.svelte
```diff
@@ -0,0 +1,175 @@
...
+      on:computedposition={positionArrow}
+    >
+      <div
+        class="tooltip"
+        class:hero={mode === 'hero'}
+        class:info={mode === 'info'}
+        class:mini={mode === 'mini'}
+        class:default={mode === 'default' || !mode}
+        transition:fade={{ duration: 60 }}
+        hidden={!visibleInternal}
```
petemill, line 90: Can we add aria attributes for hidden please

petemill: Actually maybe we don't need that because the `hidden` attribute seems to be making the content hidden from the accessibility tree

src/components/dropdown/dropdown.svelte
```diff
@@ -129,58 +131,61 @@
...
+          <Icon name="arrow-small-down" />
+        </div>
+      </slot>
+    </FormItem>
+  </div>
+  <Floating target={dropdown} placement="bottom-start">
+    <div
+      class="leo-dropdown-popup"
+      hidden={!isOpen}
+      style:min-width="{dropdown?.getBoundingClientRect().width}px"
```
petemill, line 171: This isn't getting updated when dropdown width changes. Make storybook window narrower, then click it open and popup will be too wide.

nullhook: image

Also, the content overflows. Perhaps, add shadow, rounding and border in the outerbox?

package.json
fallaciousreasoning: https://github.com/brave/reviews/issues/1400

examples/example-ui-react/src/App.tsx
fallaciousreasoning: Whoops, I was probably using it to see what colors are available :flushed:

src/components/dropdown/dropdown.svelte
fallaciousreasoning: @nullhook I'm tempted to leave that as a follow up, because I'm going on PTO next week, and it's like that even without this PR - we should definitely fix it though

petemill (approved):

fallaciousreasoning:
Blocked on security review https://github.com/brave/reviews/issues/1400

src/components/alert/alertCenter.svelte
```diff
@@ -91,7 +91,7 @@
         <div slot="actions">
           {#each alert.actions as action}
             <Button
-              kind={action.kind ?? 'filled'}
+              kind={action.kind || 'filled'}
```
petemill, line 94: cc @fallaciousreasoning this seems needed to use in projects that don't know about `??` (like brave-core's webpack v4!) as I don't think typescript sees these svelte parts, only the svelte compiler

petemill:
We can make size middleware work as a follow-up https://github.com/brave/leo/issues/391

src/components/dropdown/dropdown.stories.svelte
```diff
@@ -52,7 +52,9 @@
 </Template>
 
 <Story name="Slots" let:args>
-  <SlotInfo description="The dropdown provides several slots for customization">
+  <SlotInfo
+    description="The dropdown provides several slots for customization. It accepts 'leo-menu-item' and 'leo-option.' 'Leo-menu-item' is mainly used for context menus, which are contained within a menu, while 'leo-option' is used for selectable choices."
```
petemill, line 56: nit: a "context menu" is the whole menu. I would suggest that dropdown menu should only support leo-option, and in ButtonMenu story we can suggest both

src/components/buttonMenu/buttonMenu.stories.svelte
```diff
@@ -0,0 +1,84 @@
...
+      </div>
+    </ButtonMenu>
+  </div>
+</Template>
+
+<Story name="Default" />
+
+<Story name="Slots" let:args>
+  <SlotInfo
+    description="The ButtonMenu provides several slots for customization. It accepts 'leo-menu-item' and 'leo-option.' 'Leo-menu-item' is mainly used for context menus, which are contained within a menu, while 'leo-option' is used for selectable choices."
```
petemill, line 37:
```suggestion
    description="The ButtonMenu provides several slots for customization. It accepts '<leo-menu-item>' and '<leo-option>' elements as well as any other element types. Use '<leo-menu-item>' for menu-style commands and '<leo-option>' for select-style selectable choices."
```

nullhook: Fixed

src/components/alert/alertCenter.svelte
fallaciousreasoning: :facepalm:

PR #388, opened 2023-09-15
fix: prevent window methods being called when SSR

mattwilliams85:
Allows Leo react components to work with an SSR application

### DIFF
--- a/src/components/icon/icon.svelte
+++ b/src/components/icon/icon.svelte
@@ -21,6 +21,7 @@
     // call setIconBasePath before we go and preload the icons (or we'll get a
     // 404).
     if (isSSR) return
+
     requestIdleCallback(() => {
       const image = new Image()
       image.src = getIconUrl(lastIconBasePath, name)
--- a/src/components/icon/icon.svelte
+++ b/src/components/icon/icon.svelte
@@ -1,4 +1,5 @@
 <script context="module" lang="ts">
+  import isSSR from '../../svelteDirectives/isSSR'
   import { writable } from 'svelte/store'
 
   let lastIconBasePath = '/icons'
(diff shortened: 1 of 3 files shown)


src/components/icon/icon.svelte
```diff
@@ -14,11 +15,12 @@
 
   // Not actually used by the component, but used to preload SVGs.
   const svgCache = {}
-  export const preloadIcon = (name: string) =>
+  export const preloadIcon = (name: string) => {
     // Note: We do this in a |requestIdleCallback| because we want to do this as
     // soon as possible, but we want to make sure the consumer has a chance to
     // call setIconBasePath before we go and preload the icons (or we'll get a
     // 404).
+    if (isSSR) return
```
AlanBreck, line 23: nit: a new line may be helpful here.
```suggestion
    if (isSSR) return

```

AlanBreck (approved):
Looks good! One tiny little nit, but feel free to ignore.

src/components/icon/icon.svelte
```diff
@@ -14,11 +15,13 @@
...
+
     requestIdleCallback(() => {
```
petemill, line 25: If in nodeJS we simply ran this function immediately instead of waiting for something similar to `requestIdleCallback`, would it make for better SSR, e.g. the correct icon url would get sent to the client?

src/svelteDirectives/isSSR.ts
```diff
@@ -0,0 +1,7 @@
+export default (function isSSR(): boolean {
```
petemill, line 1: This could be used by anything, doesn't seem svelte-specific, perhaps there's a different directory name we could use.

src/components/icon/icon.svelte
fallaciousreasoning: It would, unless we hadn't configured the icon base url :laughing:

PR #389, opened 2023-09-17
CI: Run Deploy on all branches and not only "main"

petemill:
PRs like https://github.com/brave/leo/pull/383 don't have their CI steps run because they are not directly against `main`, but another branch. This fixes that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,11 +1,7 @@
 name: Deploy
 on:
   pull_request:
-    branches:
-      - main
   push:
-    branches:
-      - main
   workflow_dispatch:
 
 concurrency:


kim0:
Since the workflow has write access to PRs, this change might not be safe
cc @brave/sec-team @brave/devops-team

petemill:
@kim0 can you explain how this would be less safe than PRs against `main` considering (I believe) the same limited users have the same permissions to submit both types of branches/PR?

kim0:
> @kim0 can you explain how this would be less safe than PRs against `main` considering (I believe) the same limited users have the same permissions to submit both types of branches/PR?

yeah given all people opening PRs are trusted, I don't think this adds more risk. I'll go ahead and approve it. Sorry for the delay

kim0 (approved):

fallaciousreasoning (approved):
LGTM

kim0:
@petemill I just noticed that now any PR will run CI twice, bec it will be double counted as:
* A push to the PR's feature branch
* Also a pull_request

So I'm thinking maybe we should leave the event `push` restricted only to `main` and leave the `pull_request` event unrestricted. WDYT ?

fallaciousreasoning:
That seems more sensible!

PR #392, opened 2023-09-20
allow className on leo-menu-item for jsx

petemill:
We should probably make leo-menu-item in to a Component to allow for common menu item styles - section titles, separators, etc. In the meantime, we need to consider allowing className so that we can apply border-top

### DIFF
--- a/src/components/menu/menu.svelte
+++ b/src/components/menu/menu.svelte
@@ -7,6 +7,9 @@
           // Note: This should line up with Reacts key type, but we don't want
           // to depend on React in this layer, so we just define it manually.
           key?: string | number | null
+          // Menu items can have borders, so we need to allow that.
+          // TODO(petemill): probably make this in to a Component with props and built-in styles
+          className?: string
           children?: any
           onClick?: EventListener
         }


nullhook (approved):

src/components/menu/menu.svelte
```diff
@@ -7,6 +7,9 @@
           // Note: This should line up with Reacts key type, but we don't want
           // to depend on React in this layer, so we just define it manually.
           key?: string | number | null
+          // Menu items can have borders, so we need to allow that.
+          // TODO(petemill): probably make this in to a Component with props and built-in styles
+          className?: string
```
fallaciousreasoning, line 12: I'm not 100% sure if this will apply to the component, or if it needs to be `class`. Might be worth checking in our React example project before we merge :smile: 

Also, I think format is failing.

fallaciousreasoning (approved):
LGTM

petemill:
This was more of a test I suppose. In the end I ended up using a separate item for separator, and making sure it wasn't selectable as a menu item

PR #395, opened 2023-09-22
Menu item styles and interactivity

petemill:
Menu Item enhancements:
- Styling matches figma (hover, selection, focus)
- Pressing enter on the items actually clicks them
- Items can opt out of built-in styles and tabbing / keyboard down arrow and enter press via using a <div> (maybe it's separators or custom interactivity)
- Items can opt out of the menu immediately closing on selection by specifying `data-is-interactive={true}`

### DIFF
--- a/src/components/menu/menu.svelte
+++ b/src/components/menu/menu.svelte
@@ -41,16 +41,22 @@
   }
 
   $: menuItems = Array.from(
-    (
-      popup?.querySelector('.leo-menu-popup slot') as HTMLSlotElement
-    )?.assignedElements() ??
-      popup?.querySelectorAll('.leo-menu-popup > *') ??
+    // TODO(petemil): support slot *descendents* that are leo-menu-item or leo-option (or something else unstyled like leo-menu-action?) so that we can
+    // select items that are in complex positions (see an example in the browser app menu zoom controls).
+    (popup?.querySelector('.leo-menu-popup slot') as HTMLSlotElement)
+      ?.assignedElements()
+      ?.filter((element) =>
+        ['LEO-OPTION', 'LEO-MENU-ITEM'].includes(element.tagName)
+      ) ??
+      popup?.querySelectorAll(
+        '.leo-menu-popup > :is(leo-menu-item, leo-option'
+      ) ??
       []
   )
 
   $: {
-    for (const [menuItem, index] of menuItems.map((o, i) => [o, i] as const)) {
-      menuItem.setAttribute('tabindex', (index + 1).toString())
+    for (const menuItem of menuItems) {
+      menuItem.setAttribute('tabindex', '0')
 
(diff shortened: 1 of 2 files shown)


nullhook (approved):

petemill:
I merged this without codeowner approval due to PTO and to unblock brave-core. Let me know if anyone has extra feedback.

src/components/menu/menu.svelte
```diff
@@ -77,19 +83,28 @@
     // a change event.
     if (!item) return
 
+    // Use data-is-interactive=true to prevent the menu from closing when selected. This infers
+    // there is interacitivity inside the menu item (e.g. a Toggle), which would be good for the user
+    // to see change state and allowing the user to manually close when ready.
+    if (
```
fallaciousreasoning, line 89: Maybe worth extracting these into an array?

`const menuChildTags = ['LEO-OPTION', 'LEO-MENU-ITEM']`

src/components/menu/menu.svelte
```diff
@@ -178,50 +193,63 @@
     background: var(--leo-color-container-background);
     box-shadow: var(--leo-effect-elevation-03);
 
-    display: flex;
-    flex-direction: column;
+    // TODO(petemill): Make the "floating-ui" element be this popup element,
+    // so that we get the correct thing scrolling when overflow happens. In the meantime,
+    // overflow: 'auto' (or anything but 'visible') helps clip the content
+    // to the border-radius.
```
fallaciousreasoning, line 199: :+1:

src/components/menu/menu.svelte
```diff
@@ -77,19 +83,28 @@
     // a change event.
     if (!item) return
 
+    // Use data-is-interactive=true to prevent the menu from closing when selected. This infers
+    // there is interacitivity inside the menu item (e.g. a Toggle), which would be good for the user
+    // to see change state and allowing the user to manually close when ready.
+    if (
+      (item.tagName === 'LEO-OPTION' || item.tagName === 'LEO-MENU-ITEM') &&
+      !item.dataset.isInteractive
```
fallaciousreasoning, line 91: I think it would be ideal if we didn't have to mark items as `data-interactive=true` - it complicates the API somewhat - is it not possible to use some combination of `stopPropagation` and `preventDefault` to achieve the same result?

petemill: We still want to have `item.click()` fire. In other words, we want to be able to hit kbd arrows and then enter to fire the click event. Perhaps since we're piggybacking on element focus then "enter" might cause the element's onclick to fire? I can't remember. Overall this is probably less complicated if we make Components out of child items (since they seem to have particular design patterns) and expose attributes that affect this explicitly. Or some other way to expose options more cleanly instead of a mix of tag name, data-attributes (and class names) - data attributes everywhere?! 🙃

PR #397, opened 2023-09-25
[Menu]: Remove highlight color

fallaciousreasoning:
The `--leo-color-text-highlight` token does not exist - instead the Menu text color should not change on hover.

In addition, the Typechecking step was failing because the result of `querySelectorAll` is `Element[]` rather than `HTMLElement[]`.

**Note:** This is causing all builds to fail on `main`.


petemill (approved):
😬

Issue #404, opened 2023-09-28
Menu: do not call click handler twice when clicked


petemill:
Menu items should not call `element.click()` when clicked, only when using keyboard selection and pressing enter.

PR #405, opened 2023-09-28
Menu: fix click getting fired twice when clicking menu items

petemill:
Fix https://github.com/brave/leo/issues/404

> Menu items should not call `element.click()` when clicked, only when using keyboard selection and pressing enter.

### DIFF
--- a/src/components/buttonMenu/buttonMenu.stories.svelte
+++ b/src/components/buttonMenu/buttonMenu.stories.svelte
@@ -8,6 +8,8 @@
   import Toggle from '../toggle/toggle.svelte'
 
   export let toggleIsChecked = false
+
+  const handleAction = () => console.log('action')
 </script>
 
 <Meta title="Components/ButtonMenu" component={ButtonMenu} />
@@ -15,7 +17,8 @@
 <Template let:args>
   <div class="container">
     <ButtonMenu>
-      <leo-menu-item> Copy </leo-menu-item>
+      <!-- svelte-ignore a11y-click-events-have-key-events leo-menu-item peovides key events -->
+      <leo-menu-item on:click={handleAction}> Copy </leo-menu-item>
       <leo-menu-item> Share </leo-menu-item>
       <div class="section">Section</div>
       <leo-menu-item>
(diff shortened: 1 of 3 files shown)


nullhook (approved):

src/components/buttonMenu/buttonMenu.stories.svelte
```diff
@@ -8,14 +8,17 @@
   import Toggle from '../toggle/toggle.svelte'
 
   export let toggleIsChecked = false
```
fallaciousreasoning, line 10: Should we be exporting this from a Story? I dunno what this will actually do :laughing:

src/components/buttonMenu/buttonMenu.stories.svelte
```diff
@@ -26,12 +29,13 @@
       </leo-menu-item>
       <div class="custom-item">
         <div>Suggested questions</div>
-        <Toggle size="small" />
+        <Toggle checked={toggleIsChecked} size="small" />
```
fallaciousreasoning, line 32: Should this be 
```suggestion
        <Toggle bind:checked={toggleIsChecked} size="small" />
```
so toggling it updates the variable & vice versa?

src/components/menu/menu.svelte
```diff
@@ -85,7 +85,8 @@
 
     // Use data-is-interactive=true to prevent the menu from closing when selected. This infers
```
fallaciousreasoning, line 86:
```suggestion
    // Use data-is-interactive=true to prevent the menu from closing when selected. This impliies
```

src/components/menu/menu.svelte
```diff
@@ -102,7 +103,9 @@
       })
     }
 
-    if (item.tagName === 'LEO-MENU-ITEM') {
+    // When using keyboard navigation to call selectMenuItem, ensure click handlers happen on the item,
+    // but not when this was called in a click handler, otherwise we'll get 2x clicks.
+    if (item.tagName === 'LEO-MENU-ITEM' && e.type !== 'click') {
```
fallaciousreasoning, line 108: I'm a little lost here, but do we not call `click` on `leo-option`s too, when selecting them with the keyboard?

fallaciousreasoning (approved):
LGTM

src/components/menu/menu.svelte
petemill: No. I added a click call to leo-menu-item because we expect those to have their own click handler in the consumer, but for dropdown menus, we don't expect them to have any onClick handlers in the calling code - only Nala Dropdown code would perform the "select" action

src/components/buttonMenu/buttonMenu.stories.svelte
petemill: I don't, either

nullhook:
related but not really: do we need className to be available on leo-menu-item?

`<leo-menu-item className={} onClick={} />`

fallaciousreasoning:
I wonder if styled components only sets the `className` prop or if it also sets `class`? That's the only reason I can think of that we'd need both.

src/components/menu/menu.svelte
petemill: fixed

src/components/buttonMenu/buttonMenu.stories.svelte
petemill: good spot - fixed

fallaciousreasoning:
**Note:** I think this has broken clicking on dropdowns to close the dropdowns.

Issue #413, opened 2023-10-03
[Dropdown]: Clicking an open dropdown doesn't close the dropdown


fallaciousreasoning:
This used to work, but I think it broke with one of the latest `ButtonMenu` PRs (maybe the double click one)

cc @petemill @nullhook

petemill:
Looks like it's caused by `handleBlur` in `Menu.svelte`. Clicking the `Dropdown` element invokes that which changes `isOpen` to `false`, but then the `click` handler for `Dropdown` fires and changes `isOpen` back to `true` (via `isOpen = !isOpen`)!

PR #414, opened 2023-10-04
[Dropdown]: Make clicking an open dropdown close the dropdown

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/413


src/components/menu/menu.svelte
```diff
@@ -25,11 +25,32 @@
...
+  const dispatch = createEventDispatcher<{
+    close: CloseEventDetail
+    'select-item': SelectItemEventDetail
+  }>()
+
+  function dispatchClose(
+    originalEvent: Event,
+    reason: CloseEventDetail['reason']
+  ) {
+    if (dispatch('close', { originalEvent, reason }, { cancelable: true })) {
```
petemill, line 50: Add a comment, something like "allow the caller to cancel the default action

src/components/dropdown/dropdown.svelte
```diff
@@ -83,7 +86,13 @@
     bind:isOpen
     bind:currentValue={value}
     on:select-item={onItemSelect}
-    on:close={() => button.focus()}
+    on:close={(e) => {
+      if (e.detail.originalEvent.composedPath().includes(dropdown)) {
+        e.preventDefault()
+      } else {
+        button.focus()
+      }
```
petemill, line 94: I think it's fine but, again, needs a comment especially here about why we need to do it

petemill (approved):

PR #418, opened 2023-10-05
Fix Tailwind font token output

AlanBreck:
This PR excludes `macos` and `windows` font styles from the Tailwind output and removes the `default-` prefix from TW font token names.

Related to the above, this also fixes the regex in `leoParser` to exclude emojis.


package-lock.json
fallaciousreasoning, line 1: FWIW, I think this change is fine - we probably forgot to do an `npm i` after pinning the package versions.

fallaciousreasoning (approved):
LGTM

Issue #419, opened 2023-10-05
ButtonMenu should get it's button from slot


petemill:
Let's stop providing `<button>` inside ButtonMenu, and have the consumers provide their own button-y thing. We can put an on-click on the slot / parent, and the click on the button-y thing will bubble up to our handler.

\* button-y = button-like thing

Issue #416, opened 2023-10-04
[ButtonMenu]: Tab navigation doesn't work quite right


fallaciousreasoning:
There seem to be two elements which take focus in the close button menu:

1. Focus element before the button menu
2. Press tab - ButtonMenu is focused. Pressing enter will open/close it
3. Press tab - ButtonMenu is focused. Pressing enter will open/close it (I think this is actually the anchor element of the buttonmenu)
4. Press tab - next element is focused

cc @nullhook

PR #420, opened 2023-10-06
[ButtonMenu]: Fix double focus issue with buttons

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/416
Fixes https://github.com/brave/leo/issues/419

Previously, if you used a focusable element as the anchor of the `ButtonMenu` then the button menu would accept two tabs to focus:
1. Focus the outer button, from the menu component
2. Focus the inner anchor-content element

This PR takes advantage of event bubbling, to handle clicks from the anchor element (which can be triggered via the keyboard with space or enter when the button is focused).

**Breaking:** This will break usages of the `ButtonMenu` where the `anchor-content` was not focusable. However, as the `ButtonMenu` doesn't have any focusable affordances, this is hardly worse than what we have currently.


fallaciousreasoning:
@petemill PTAL

AlanBreck (approved):

petemill (approved):
ok but let's fix AI Chat when this merges to brave-core. There's a ButtonMenu in the main-feature-menu/index.tsx and in the conversation list item.

 cc @nullhook as maybe this fixes the alignment issue you're assigned to anyway

PR #424, opened 2023-10-10
Set z-index for alert center

sangwoo108:
Currently, alert UI could be below other UI components. Sets the default z-index and makes a css variables to control the z-index

### Before
before

### After
after


sangwoo108:
Hi, folks. I was trying to use this awesome alert center in Playlist but find it stacked below other UI. Could you take a look?

fallaciousreasoning (approved):
LGTM!

Issue #407, opened 2023-09-28
Figure out how we can have compile errors when react is using the wrong event name


petemill:

benjaminknox:
I think I have a solution based for this here: https://github.com/brave/leo/pull/440

If ☝🏻 isn't what you're looking for let me know and I don't mind trying again

PR #440, opened 2023-10-17
[Component]: adds event types to react

benjaminknox:
Resolves https://github.com/brave/leo/issues/407

For react components, currently  generic typing allows any event prop to be used (e.g. `<LeoButton onClick={...}>`), if an incorrect event name is used (e.g. `<LeoButton onClicked={...}>`) there is no type error. Ideally a type error for an incorrect prop name would show.

This solves this by generating an event type for the react components by parsing the corresponding svelte type file, it should contain event types resulting from the typed `createEventDispatcher`. When there are no events defined in the svelte file it doesn't include any event types.

For example, in [input.svelte#L74-L84](https://github.com/brave/leo/blob/7a65a548c3c801fcdf8b86a98118a758bfa004c8/src/components/input/input.svelte#L74-L84):

```ts
  type InputEventDetail = {
    innerEvent: Event & { target: HTMLInputElement }
    value: string
    valueAsNumber: number
    valueAsDate: number
  }

  const dispatch = createEventDispatcher<{
    change: InputEventDetail
    input: InputEventDetail
    focus: InputEventDetail
    blur: InputEventDetail
    keydown: InputEventDetail
    keyup: InputEventDetail
    keypress: InputEventDetail
    focusin: InputEventDetail
    focusout: InputEventDetail
  }>()
```

Results in a type definition that looks like this:
```ts
import type * as React from 'react'
import type { ReactProps } from '../src/components/svelte-react'
import type { InputEvents as SvelteEvents, InputProps as SvelteProps } from '../types/components/input/input.svelte';

export type InputEventProps = {
  onChange?: (event: CustomEvent<{
      innerEvent: Event & {
          target: HTMLInputElement;
      };
      value: string;
      valueAsNumber: number;
      valueAsDate: number;
  }) => void;
  onInput?: (event: CustomEvent<{...}>) => void;
  onFocus?: (event: CustomEvent<{...}>) => void;
  onBlur?: (event: CustomEvent<{...}>) => void;
  onKeydown?: (event: CustomEvent<{...}>) => void;
  onKeyup?: (event: CustomEvent<{...}>) => void;
  onKeypress?: (event: CustomEvent<{...}>) => void;
  onFocusin?: (event: CustomEvent<{...}>) => void;
  onFocusout?: (event: CustomEvent<{...}>) => void
}

export type InputProps = InputEventProps & ReactProps<SvelteProps, SvelteEvents>;
export default function InputReact(props: React.PropsWithChildren<InputProps>): JSX.Element

// As we don't currently have type definitions for the web components, export
// the Type Definitions from the Svelte component.
export * from '../types/components/input/input.svelte'
```

https://github.com/brave/leo/assets/5668789/e27cbc1e-7a9c-4162-9710-21fdce5a46ed

### DIFF
--- a/src/components/button/button.svelte
+++ b/src/components/button/button.svelte
@@ -50,7 +50,9 @@
 
   $: tag = href ? 'a' : ('button' as 'a' | 'button')
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    click: {}
+  }>()
 
   /**
    * Optional click handler
--- a/src/components/checkbox/checkbox.svelte
+++ b/src/components/checkbox/checkbox.svelte
@@ -22,7 +22,10 @@
   export let isDisabled = false
   export let size: Sizes = 'normal'
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    change: { checked: boolean }
+  }>()
+
   function change(e) {
     dispatch('change', {
       checked: e.target.checked
--- a/src/components/collapse/collapse.svelte
+++ b/src/components/collapse/collapse.svelte
@@ -11,7 +11,10 @@
   export let isOpen: boolean | undefined
   $: isOpenInternal = isOpen ?? false
 
-  const dispatcher = createEventDispatcher()
+  const dispatcher = createEventDispatcher<{
+    toggle: { open: boolean }
+  }>()
+
   const toggle = (e) => {
     e.preventDefault()
 
--- a/src/components/dialog/dialog.svelte
+++ b/src/components/dialog/dialog.svelte
@@ -13,7 +13,10 @@
   export let backdropClickCloses = true
   export let animate = true
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    close: {}
+    back: {}
+  }>()
 
   let dialog: HTMLDialogElement
   $: {
--- a/src/components/dropdown/dropdown.svelte
+++ b/src/components/dropdown/dropdown.svelte
@@ -28,10 +28,11 @@
   export let size: Size = 'normal'
   export let required = false
   export let mode: Mode = 'outline'
-
   export let showErrors = false
 
-  let dispatch = createEventDispatcher()
+  let dispatch = createEventDispatcher<{
+    change: { value: string }
+  }>()
 
   let isOpen = false
   let button: HTMLButtonElement
--- a/src/components/radioButton/radioButton.svelte
+++ b/src/components/radioButton/radioButton.svelte
@@ -22,7 +22,9 @@
   export let isDisabled = false
 
   const tagName = 'leo-radiobutton'
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    change: { value: string | number | any }
+  }>()
 
   function changed(e) {
     if (isDisabled || !e.target.checked) return
--- a/src/components/svelte-react.ts
+++ b/src/components/svelte-react.ts
@@ -6,7 +6,8 @@ import {
   useCallback,
   type ForwardedRef,
   type PropsWithChildren,
-  useState
+  useState,
+  type DOMAttributes
 } from 'react'
 import type { SvelteComponentTyped } from 'svelte'
 
@@ -35,8 +36,6 @@ export type SvelteEvents<T> = T extends SvelteComponentTyped<
   ? Events
   : {}
 export type ReactProps<Props, Events> = Props & {
-  [P in keyof Events as `on${Capitalize<P & string>}`]?: (e: Events[P]) => void
-} & {
   ref?: ForwardedRef<Partial<Props & HTMLElement> | undefined>
 } & {
   // Note: The div here isn't important because all props in intrinsicProps are
@@ -45,6 +44,16 @@ export type ReactProps<Props, Events> = Props & {
   [P in IntrinsicProps]?: JSX.IntrinsicElements['div'][P]
 }
 
+type EventPropsNames = keyof DOMAttributes<unknown>
+
+type EventPropsNameMap = {
+  [P in EventPropsNames as Lowercase<P>]: P
+}
+export type EventProps<T> = {
+  [P in keyof T as P extends Lowercase<EventPropsNames>
+    ? EventPropsNameMap[P]
+    : P]: T[P]
+}
 const useEventHandlers = (props: any) => {
   const [el, setEl] = useState<HTMLElement>()
   const lastValue = useRef<{ [key: string]: (...args: any[]) => any }>({})
--- a/src/components/svelte-react.ts
+++ b/src/components/svelte-react.ts
@@ -52,6 +52,8 @@ type EventPropsNameMap = {
 export type EventProps<T> = {
   [P in keyof T as P extends Lowercase<EventPropsNames>
     ? EventPropsNameMap[P]
+    : P extends `on${infer EventName}`
+    ? `on${Capitalize<EventName>}`
     : P]: T[P]
 }
 const useEventHandlers = (props: any) => {
--- a/src/components/toggle/toggle.svelte
+++ b/src/components/toggle/toggle.svelte
@@ -18,7 +18,10 @@
   let dragStartX: number | undefined
   let dragOffsetX: number = 0
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    change: { checked: boolean }
+  }>()
+
   const onChange = (newValue?: boolean) => {
     if (newValue === undefined) newValue = !checked
     checked = newValue
--- a/src/components/tooltip/tooltip.svelte
+++ b/src/components/tooltip/tooltip.svelte
@@ -33,7 +33,9 @@
   // controlled and uncontrolled states for this component.
   $: visibleInternal = visible ?? false
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    visibilitychange: { visible: boolean }
+  }>()
 
   let tooltip: HTMLElement
   let arrow: HTMLElement
--- a/src/scripts/gen-react-bindings.js
+++ b/src/scripts/gen-react-bindings.js
@@ -8,6 +8,50 @@ fs.mkdir(REACT_BINDINGS_DIRECTORY, { recursive: true })
 const COMPONENT_PREFIX = 'leo'
 const SVELTE_REACT_WRAPPER_PATH = '../shared/svelte-react.js'
 
+const getEventsTypeDefinition = (
+  componentName,
+  componentEventNames,
+  svelteFilePath
+) => `
+export type ${componentName}EventProps = EventProps<{
+${componentEventNames
+  .map(
+    ({ eventName, detailType }) =>
+      `  ${eventName}?: (event: ${detailType}) => void`
+  )
+  .join(';\n')}
+}>
+`
+
+const findEventsTypeDefinition = async (svelteFilePath, componentName) => {
+  const pathToType = path.join(
+    __dirname,
+    '../../',
+    'types',
+    path.relative('./src', svelteFilePath + '.d.ts')
+  )
+  const fileContents = await fs.readFile(pathToType)
+
+  const typeRegex = /\s+([a-zA-Z0-9]+): (CustomEvent<{(\n|.)*?}>)/gm
+
+  const componentEventNames = [
+    ...fileContents.toString().matchAll(typeRegex)
+  ].map((match) => ({
+    eventName: `on${match[1]}`,
+    detailType: match[2]
+  }))
+
+  if (componentEventNames.length > 0) {
+    return getEventsTypeDefinition(
+      componentName,
+      componentEventNames,
+      svelteFilePath
+    )
+  } else {
+    return []
+  }
+}
+
 const getComponentGenerics = async (svelteFilePath, componentName) => {
   const relativePath = path.relative('./src/components', svelteFilePath)
   const typingsPath = path.join('./types/components', relativePath) + '.d.ts'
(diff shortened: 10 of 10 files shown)


src/scripts/gen-react-bindings.js
```diff
@@ -79,8 +116,10 @@ export * from '../web-components/${fileNameWithoutExtension}.js'
 import type * as React from 'react'
 import type { ReactProps } from '../src/components/svelte-react'
 import type { ${componentName}Events as SvelteEvents, ${componentName}Props as SvelteProps } from '../types/components/${containingFolder}/${fileName}';
-
-export type ${componentName}Props${funcConstraints} = ReactProps<SvelteProps${propParams}, SvelteEvents${propParams}>;
+${eventTypeDefinition && eventTypeDefinition}
+export type ${componentName}Props${funcConstraints} = ${
+    eventTypeDefinition && `${componentName}EventProps &`
+  } ReactProps<SvelteProps${propParams}, SvelteEvents${propParams}>;
```
benjaminknox, line 122: This is where the type definition is generated.

src/scripts/gen-react-bindings.js
```diff
@@ -8,6 +9,38 @@ fs.mkdir(REACT_BINDINGS_DIRECTORY, { recursive: true })
...
 const SVELTE_REACT_WRAPPER_PATH = '../shared/svelte-react.js'
 
+const getEventsTypeDefinition = (componentName, componentEventNames) => {
+  if (componentEventNames.length === 0) return null
+
+  return `
+export type ${componentName}EventProps = {
+${componentEventNames.map((event) => `  ${event}?: (event: CustomEvent) => void`).join(';\n')}
+}
+`
```
benjaminknox, line 19: Creates a type that gets associated with the component type during the generation of the react components.

package.json
```diff
@@ -9,6 +9,7 @@
     "@floating-ui/dom": "1.4.4",
     "@tsconfig/svelte": "3.0.0",
     "lodash.camelcase": "4.3.0",
+    "lodash.capitalize": "^4.2.1",
```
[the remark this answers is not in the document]
fallaciousreasoning: In general, we prefer to avoid adding new deps, if we can :smile: If you don't mind, could you just do it in Vanilla?

src/components/button/button.svelte
```diff
@@ -49,7 +49,9 @@
 
   $: tag = href ? 'a' : ('button' as 'a' | 'button')
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    click: CustomEvent
```
fallaciousreasoning, line 53: I think this is actually the type of the `detail` property on the custom event (i.e. `e.detail`, so it should be fine to type this as `{}`
```suggestion
    click: {}
```

src/components/checkbox/checkbox.svelte
```diff
@@ -22,7 +22,10 @@
   export let isDisabled = false
   export let size: Sizes = 'normal'
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    change: CustomEvent
```
fallaciousreasoning, line 26:
```suggestion
    change: { checked: boolean }
```

src/components/collapse/collapse.svelte
```diff
@@ -11,7 +11,10 @@
   export let isOpen: boolean | undefined
   $: isOpenInternal = isOpen ?? false
 
-  const dispatcher = createEventDispatcher()
+  const dispatcher = createEventDispatcher<{
+    toggle: CustomEvent
```
fallaciousreasoning, line 15:
```suggestion
    toggle: { open: boolean }
```

src/components/dialog/dialog.svelte
```diff
@@ -13,7 +13,9 @@
   export let backdropClickCloses = true
   export let animate = true
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    close: CustomEvent
```
fallaciousreasoning, line 17:
```suggestion
    close: {}
```

src/components/dropdown/dropdown.svelte
```diff
@@ -29,7 +29,9 @@
   export let required = false
   export let mode: Mode = 'outline'
 
-  let dispatch = createEventDispatcher()
+  let dispatch = createEventDispatcher<{
+    change: CustomEvent
```
fallaciousreasoning, line 33:
```suggestion
    change: { value: string }
```

src/components/radioButton/radioButton.svelte
```diff
@@ -22,7 +22,9 @@
   export let isDisabled = false
 
   const tagName = 'leo-radiobutton'
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    change: CustomEvent
```
fallaciousreasoning, line 26:
```suggestion
    change: { value: string | number | any }
```

src/components/toggle/toggle.svelte
```diff
@@ -18,7 +18,10 @@
   let dragStartX: number | undefined
   let dragOffsetX: number = 0
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    change: CustomEvent
```
fallaciousreasoning, line 22:
```suggestion
    change: {}
```

src/components/tooltip/tooltip.svelte
```diff
@@ -25,7 +25,9 @@
   // controlled and uncontrolled states for this component.
   $: visibleInternal = visible ?? false
 
-  const dispatch = createEventDispatcher()
+  const dispatch = createEventDispatcher<{
+    visibilitychange: CustomEvent
```
fallaciousreasoning, line 29:
```suggestion
    visibilitychange: { visible: boolean }
```

src/scripts/gen-react-bindings.js
```diff
@@ -1,4 +1,5 @@
 const fs = require('fs/promises')
+const capitalize = require('lodash.capitalize')
```
fallaciousreasoning, line 2: nit: just do this in vanilla js

src/scripts/gen-react-bindings.js
```diff
@@ -8,6 +9,40 @@ fs.mkdir(REACT_BINDINGS_DIRECTORY, { recursive: true })
 const COMPONENT_PREFIX = 'leo'
 const SVELTE_REACT_WRAPPER_PATH = '../shared/svelte-react.js'
 
+const getEventsTypeDefinition = (componentName, componentEventNames) => {
+  if (componentEventNames.length === 0) return null
```
fallaciousreasoning, line 13: Might be worth logging a warning if we have an untyped `createEventDispatcher`

src/scripts/gen-react-bindings.js
```diff
@@ -8,6 +9,40 @@ fs.mkdir(REACT_BINDINGS_DIRECTORY, { recursive: true })
...
+
+  const eventDispatcherRegex = /createEventDispatcher<{((\n.*)+)\s}>/g
+  const findDefinedEvents = fileContents
+    .toString()
+    .match(eventDispatcherRegex)?.[0]
+
+  if (findDefinedEvents) {
+    const eventNameRegex = /(\w+)\:.*/g
+
+    const componentEventNames = [
```
fallaciousreasoning, line 36: We can actually type the event as `CustomEvent<DetailType>` too

fallaciousreasoning (changes requested):
Wow thanks for this @benjaminknox - I'm going to take a couple of days to mull it over (and I'm away the rest of this week) but this seems like a sensible approach.

The main thing we might be missing is the ability to accept standard HTML events. When we wrap these in a WebComponent, it actually supports all the standard events (onClick, onMouseEnter, onMouseDown, onKeyDown ect), so we should make sure we are still able to accept those events too

package.json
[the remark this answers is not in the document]
benjaminknox: Yeah, no problem, done

src/components/button/button.svelte
benjaminknox: I updated the code to do it this way, I initially used the `e.detail` property actually, no problem

Just a note about this though: The svelte code makes it seem like it is the `event.detail` property that should end up being passed in the event callback, I noticed in the resulting react components the `CustomEvent` event object is passed at runtime though.

benjaminknox:
> The main thing we might be missing is the ability to accept standard HTML events. When we wrap these in a WebComponent, it actually supports all the standard events (onClick, onMouseEnter, onMouseDown, onKeyDown ect), so we should make sure we are still able to accept those events too

Thanks @fallaciousreasoning, I think I have a possible solution for this, the details are in a PR I created against this branch in my fork of leo: https://github.com/benjaminknox/leo/pull/1

src/scripts/gen-react-bindings.js
benjaminknox: Done, to correctly define the type I rewrote this function to do a regex on the `/types/*.svelte.d.ts` file.

I had to rewrite it this way because in some svelte files the `DetailType` was defined separately, like in [menu.svelte](https://github.com/brave/leo/blob/298a59f8f111a0abbc2fc9b22885ca59e17d0b15/src/components/menu/menu.svelte#L28) but the svelte type file has the event detail types without the separate definition.

src/scripts/gen-react-bindings.js
benjaminknox: I am not warning because I rewrote this to use the `/types/*.svelte.d.ts` file, so the script doesn't know if `createEventDispatcher` is typed anymore unfortunately

src/components/button/button.svelte
fallaciousreasoning: Interesting - the type you pass in here is the type of `event.detail`, because the type of the eventArg is `CustomEvent<DetailType>`

https://github.com/sveltejs/svelte/blob/522d06a6add1de1779a87c83bee0a43163044c75/packages/svelte/src/runtime/internal/lifecycle.js#L83

It might be that we're handling this typing wrong with the `gen-react-types` code though :disappointed:

benjaminknox: After working on this a bit more I think the way this is handled makes sense, there's just some semantics that I initially found confusing:

What I initially expected:
1. Define the event detail type in the `createEventDispatcher<{ event: CustomEvent<DetailType> }>` call in svelte
2. The event handler in react passes `CustomEvent<DetailType>`

How it actually works:
1. Define the event detail type in the `createEventDispatcher<{ event: DetailType }>` call in svelte
2. The event handler in react passes `CustomEvent<DetailType>`

I think this is fine

benjaminknox:
@fallaciousreasoning I think I've addressed all of your comments here.

Although there's one more thing bothering me, we end up with event names that are not quite camelcase, like in the `Input` component in the example above there is `onKeypress` instead of the react `onKeyPress`.

Is this ok to keep this way?

I have a few ideas on how to fix this:
1. We could implement something like [this library](https://www.npmjs.com/package/wordsninja) to parse the words in the event name.
2. We could limit it to the events that are standard on react elements and match those event names case insensitively, e.g. `onKeydown => onKeyDown`, `onMouseleave => onMouseLeave`, or `onFocusout => onFocusOut`. (leaving any non-standard events)

Any thoughts?

fallaciousreasoning:
Hmm, you're right, that's not ideal - shall we see if we can solve it? I think I'd be pretty happy with a solution which just handles the standard HTML events, which we can probably do as a Typescript-only fix:

image
(my super hacky PoC)

```ts
type EventNames = 'onMouseDown' | 'onMouseEnter' | 'onClick' | 'onKeyDown' | 'onKeyUp'
type EventNameMap = {
  [P in EventNames as Lowercase<P>]: P
}

type Events<T> = {
  [P in keyof T as (P extends Lowercase<EventNames> ? EventNameMap[P] : P)]: T[P]
} 

let f: Events<{ 'onclick': () => {}, 'ondidthething': () => {}, onmouseenter: () => {} }>
f.<CTRL+Space>
```

src/components/svelte-react.ts
```diff
@@ -45,6 +44,16 @@ export type ReactProps<Props, Events> = Props & {
   [P in IntrinsicProps]?: JSX.IntrinsicElements['div'][P]
 }
 
+type EventPropsNames = keyof DOMAttributes<unknown>
```
benjaminknox, line 47: I was able to get a list of the props from react here, @fallaciousreasoning using your prototype this actually works so standard react event prop names like `onKeyDown` are valid.

fallaciousreasoning:
@benjaminknox I finally got around to testing this out in https://github.com/brave/brave-core, which is our main consumer of this repo. This is mostly working great but I did notice a couple of things:

1. On the `LeoButton` component, the event names which are forwarded to the internal button are named `on:<eventname>` instead of `on<Eventname>`
2. We seem to have lost the uppercase letter at the start of the event name (i.e. `onClose` is now `onclose`)

I think if we fix these it should be good to land :smile: 


for capitalizing `on` in front of event names I think this should work:

```ts
type Events = ['onclick', 'onclose', 'onopen', 'onchange', 'oninput']

type CasedBetter<T> = T extends `on${infer K}` ? `on${Capitalize<K>}` : never

type CapitalOn = CasedBetter<Events[number]>
// type CapitalOn = "onClick" | "onClose" | "onOpen" | "onChange" | "onInput"
```

fallaciousreasoning:
@benjaminknox I pushed a commit  to fix the event names :smile:

fallaciousreasoning (approved):
This LGTM - if you sign the commits I can land it :smile:

benjaminknox:
> This LGTM - if you sign the commits I can land it 😄

Done! 

> We seem to have lost the uppercase letter at the start of the event name (i.e. onClose is now onclose)

Good catch thanks @fallaciousreasoning

fallaciousreasoning:
Thanks heaps for this @benjaminknox!

PR #450, opened 2023-10-26
[Navigation]: adds Navigation component

AlanBreck:


src/components/navigation/navigationHeader.svelte
```diff
@@ -0,0 +1,14 @@
+<header class="leo-navigation-header">
```
fallaciousreasoning, line 1: I don't know if this should use the `header` element - it isn't necessarily going to contain the header for, say an article, but rather for the site (with the nav links).

Don't know well enough to say either way :laughing:

src/components/navigation/navigationItem.svelte
```diff
@@ -0,0 +1,86 @@
...
+<li class="leo-navigation-item">
+  <a {href} {...$$restProps} class:isActive>
+    {#if icon}
+      <Icon name={icon} />
+    {/if}
+    <slot />
+  </a>
+
+  {#if $$slots.subnav}
+    <slot name="subnav" />
```
fallaciousreasoning, line 28: :+1:

src/components/navigation/navigationItem.svelte
```diff
@@ -0,0 +1,86 @@
+<script lang="ts">
+  import type { SvelteHTMLElements } from 'svelte/elements'
+  import type { IconName } from '../../../icons/meta'
+  import Icon from '../icon/icon.svelte'
+
+  type Href = $$Generic<string | undefined>
```
fallaciousreasoning, line 6: I suspect this doesn't need to be `Generic` - it can just be `string | undefined`.

fallaciousreasoning: In fact, I don't think we need the `href` for anything else, so you can just use the default props from the `a` element, and remove all references to the `href` in this class :smile:

src/components/navigation/navigationItem.svelte
```diff
@@ -0,0 +1,86 @@
...
+
+  type $$Props = Omit<SvelteHTMLElements['a'], 'class' | 'href'> & {
+    href: Href
+    icon?: IconName
+    isActive?: boolean
+  }
+
+  export let href: Href
+  export let icon: IconName = undefined
+  export let isActive: boolean = false
```
fallaciousreasoning, line 16: Maybe this could be
```suggestion
  export let isActive?: boolean = window.location.endsWith(href)
```

or something. It'd be nice to do some rudimentary matching to make the component easier to use :smile:

src/components/navigation/navigationItem.svelte
```diff
@@ -0,0 +1,86 @@
...
+          background: var(--leo-color-text-interactive);
+          position: absolute;
+          left: 0;
+          top: 50%;
+          transform: translateY(-50%);
+        }
+      }
+    }
+
+    :global(ul) {
```
fallaciousreasoning, line 80: Note: This won't work with `web-components`. Maybe try:

`:global(ul), ::slotted(ul) {`

You can test it in the `plain-html` example, to double check it works. Sometimes it's a bit fiddly

fallaciousreasoning (commented):
Looks nice!

src/components/navigation/navigationItem.svelte
AlanBreck: In order for this to work, though, we'd have to have a listener for the component to update when `window.location` changes in the event that we're using client-side navigation.

fallaciousreasoning: Probably worth adding - I can't imagine when you wouldn't want this to work that way by default? Otherwise the parent app will need to manage it

AlanBreck: I'm not seeing means of doing this with VanillaJS that will work broadly enough. There's the `popstate` event, but it only triggers in specific cases, and SvelteKit's client side navigation does not trigger the event.

fallaciousreasoning: I think the normal way of doing this is patching the `pushState` & `replaceState`, as well as listening to `popstate`.

```tsx

const onChange = () => {/* do stuff*}

['pushState', 'replaceState'].forEach((name) => {
  const original = history[name]
  history[name] = function() {
    original.apply(history, arguments)
    onChange()
  }
})
```

AlanBreck: But that would be up to the consuming app, and not Nala, yes?

fallaciousreasoning: I mean, there's not really any reason we can't do it in Nala is there? Externally (i.e. in the consuming app) nothing will change except the NavigationLinks will Just Work(tm).

Not super important, so we can leave it out if you think it's going to make this PR too complicated.

src/components/icon/icon.svelte
```diff
@@ -22,7 +22,7 @@
     // 404).
     if (isSSR) return
 
-    requestIdleCallback(() => {
+    window.requestIdleCallback?.(() => {
```
AlanBreck, line 25: Sneaking a few extra things into this PR. Can move into separate PRs if needed.

@fallaciousreasoning, `requestIdleCallback` is not supported in Safari. Wonder if we should let it go (above approach), or check if it exists, and use an alternative if not found.

fallaciousreasoning: Booooo Safari. What a pain. Probably fine to just not do anything on Safari - this can just stay as a progressive enhancement.

src/components/navigation/navigationHeader.svelte
AlanBreck: From seeing it's use, I believe the intent is that this would be the `header` for the page.

src/components/navigation/navigationItem.svelte
AlanBreck: What if the item is a button, though?

src/components/icon/icon.svelte
AlanBreck: Ok, so leave as is above, yes?

src/components/navigation/navigationItem.svelte
```diff
@@ -0,0 +1,150 @@
...
+  export let isLoading: boolean = false
+  export let isDisabled: boolean = false
+  export let isCurrent: boolean = window.location.pathname === href
+  export let inList: boolean = true
+
+  const checkIfCurrent = () => {
+    isCurrent = window.location.pathname === href
+  }
+
+  $: tag = href ? 'a' : ('button' as 'a' | 'button')
```
AlanBreck, line 55: This is inexplicably broken when using with web components.

AlanBreck: Broken because custom elements doesn't seem to like the double `<svelte:elements>` bit.

fallaciousreasoning: Svelte is a bit funky :/

src/components/navigation/navigation.stories.svelte
```diff
@@ -0,0 +1,102 @@
...
+  import SlotInfo from '../../storyHelpers/SlotInfo.svelte'
+  import Slot from '../../storyHelpers/Slot.svelte'
+  import Toggle from '../toggle/toggle.svelte'
+</script>
+
+<Meta
+  title="Components/Navigation"
+  component={Navigation}
+  argTypes={{
+    kind: {
```
fallaciousreasoning, line 20: We should probably list all the args here (including CSS props)

src/components/navigation/navigation.stories.svelte
```diff
@@ -0,0 +1,102 @@
...
+      <Icon slot="logo" name="social-brave-release-favicon-fullheight-color" />
+      <h1><span class="logo-mark">Brave</span> Accounts</h1>
+    </NavigationHeader>
+
+    <NavigationMenu>
+      {#each [1, 2, 3, 4, 5] as item}
+        <NavigationItem
+          href="#"
+          icon="browser-ntp-widget"
+          isCurrent={item === 3}>Item {item}</NavigationItem
```
fallaciousreasoning, line 41: We could include this in the `args`, so you can change the current item

src/components/navigation/navigation.stories.svelte
```diff
@@ -0,0 +1,102 @@
...
+    </NavigationActions>
+  </Navigation>
+</Template>
+
+<Story name="Slots" let:args>
+  <SlotInfo description="The navigation component has the following slots:">
+    <Slot
+      name="default"
+      explanation="This default slot should contain a NavigationMenu component, which in turn can contain NavigationItem and NavigationSeparator components."
+    ></Slot>
```
fallaciousreasoning, line 77: Maybe worth adding examples of the items in these slots

src/components/navigation/navigation.stories.svelte
```diff
@@ -0,0 +1,102 @@
...
+                Testing
+              </NavigationItem>
+            </NavigationMenu>
+          </NavigationItem>
+        </NavigationMenu>
+      </NavigationItem>
+    </NavigationMenu>
+
+    <NavigationActions slot="actions">
+      <Toggle>
```
fallaciousreasoning, line 64: Okay, in the story this looks kinda odd because it's just sitting there by itself - should it look more connected to the rest of the menu?

examples/plain-html/index.html
```diff
@@ -176,6 +183,42 @@
         <leo-button kind="filled">Primary</leo-button>
       </div>
     </leo-dialog>
+
+    <leo-navigation>
+      <leo-navigationheader slot="header">
+        <leo-icon slot="logo" name="social-brave-release-favicon-fullheight-color"></leo-icon>
+        <h1 style="font: var(--leo-font-primary-heading-h4)"><span class="logo-mark">Brave</span> Accounts</h1>
```
fallaciousreasoning, line 190:
```suggestion
        <h1 style="font: var(--leo-font-heading-h4)"><span class="logo-mark">Brave</span> Accounts</h1>
```

that token has been removed!

src/components/navigation/navigation.stories.svelte
```diff
@@ -0,0 +1,102 @@
...
+      explanation="This should contain the NavigationAction component, and is used to hold action items like theme switcher, profile avatar, etc"
+    ></Slot>
+  </SlotInfo>
+</Story>
+
+<Story name="Primary" let:args />
+
+<style>
+  h1 {
+    font: var(--leo-font-primary-heading-h4);
```
fallaciousreasoning, line 95:
```suggestion
    font: var(--leo-font-heading-h4);
```

fallaciousreasoning: Should we automatically set this for the `NavigationMenu`?

src/components/navigation/navigation.svelte
```diff
@@ -0,0 +1,34 @@
+<script lang="ts">
+  export let kind: 'vertical' | 'horizontal' = 'vertical'
+</script>
+
+<div
+  class="leo-navigation"
+  class:isVertical={kind === 'vertical'}
```
fallaciousreasoning, line 7: optional:

```suggestion
  class:is-vertical={kind === 'vertical'}
```

or 
```suggestion
  class="leo-navigation {kind}"
```

we change between camelCase and kebab-case everywhere so this is 100% optional :laughing:

src/components/navigation/navigationItem.svelte
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * Optional click handler
+   */
+  const dispatch = createEventDispatcher()
+  function onClick(event) {
+    dispatch('click', event)
+  }
+
+  onMount(() => {
+    ;['pushState', 'replaceState'].forEach((name) => {
```
fallaciousreasoning, line 66: :heart: thanks heaps for doing this @AlanBreck!

src/components/navigation/navigationItem.svelte
```diff
@@ -0,0 +1,152 @@
...
+      const original = history[name]
+      history[name] = function () {
+        original.apply(history, arguments)
+        checkIfCurrent()
+      }
+    })
+
+    window.addEventListener('popstate', checkIfCurrent)
```
fallaciousreasoning, line 74: Might want to listen to `hashchange` too? Do we need to clean these up when the component gets unmounted too?

src/components/navigation/navigationMenu.svelte
```diff
@@ -0,0 +1,25 @@
+<script lang="ts">
+  let isSubnav = $$props.slot === 'subnav'
```
fallaciousreasoning, line 2: ooooh nice! Does this work in WebComponents land?

src/components/navigation/navigationSeparator.svelte
fallaciousreasoning, line 1: Optional: This almost feels like it could be it's own `Hr` component? Surely we'll use this in other places too?

fallaciousreasoning (commented):
LGTM

src/components/navigation/navigation.stories.svelte
AlanBreck: The navigation component itself doesn't have any CSS props. Some of the sub components have more props/CSS props, but I'm not sure that we'd want to add them here?

fallaciousreasoning: Hmmm, that's a good point. It'd be good to have some way to figure out all the different knobs for the component, but I guess we can do that separately.

src/components/navigation/navigationItem.svelte
fallaciousreasoning: Hmm okay. It does make me feel a little weird landing it without working WC support, as a lot of the usages for these components are in Brave Core, which is solely WC based. 

That said, I want to unblock you, so I guess we can land as is, if you're okay to come back and try and fix it when you have some time?

I'd prefer to avoid a situation where we have a bunch of functionality which only works in Svelte, and a bunch which only works in WCs.

AlanBreck: I'm investigating a bit further, and it's almost as if the reactivity of `href` is lost when nested.

PR #451, opened 2023-10-26
Input fixes

AlanBreck:
- adds more support for default attributes on Input component

Currently, the input component doesn't support `type="email"`, this PR adds support for that and other input types which we may want to use.

- fixes background color on formControl and renames classes

The primary fix here is an incorrect CSS var, but I've additionally renamed the classes to follow suit with the Button component: `isFilled`, `isOutline`, etc. This is helpful especially in the context of projects that use the extracted component styles as Tailwind plugins since `outline` is a default class to style the `outline` property on interactive elements.

### DIFF
--- a/src/components/formItem/formItem.svelte
+++ b/src/components/formItem/formItem.svelte
@@ -54,11 +54,11 @@
 
 <label
   class="leo-control"
-  class:small={size === 'small'}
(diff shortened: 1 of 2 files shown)


fallaciousreasoning (approved):
LGTM

PR #453, opened 2023-10-26
splits tailwind component plugins into individual imports

AlanBreck:
This makes it so that Tailwind component plugins (extracted from the Svelte components) can be imported individually into a project's Tailwind config instead of all at once.

### DIFF
--- a/src/tokens/transformation/tailwind/extractComponentStyles.js
+++ b/src/tokens/transformation/tailwind/extractComponentStyles.js
@@ -1,6 +1,5 @@
 const svelte = require('svelte/compiler')
-const { readdirSync, statSync } = require('fs')
-const { writeFile, readFile } = require('fs/promises')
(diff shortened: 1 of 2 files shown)


src/tokens/transformation/tailwind/extractComponentStyles.js
```diff
@@ -46,7 +45,11 @@ module.exports = {
           const fnName = fileParts.name === 'link' ? 'addBase' : 'addComponents'
 
           await writeFile(
-            join(config.buildPath, 'plugins', `${fileParts.name}Component.js`),
+            join(
+              config.buildPath,
+              'plugins/components',
```
fallaciousreasoning, line 50: Note: the build seems to be failing

fallaciousreasoning (approved):
LGTM, but the build seems to be failing

AlanBreck:
> LGTM, but the build seems to be failing

Fixed.

Issue #455, opened 2023-10-26
Form item error management should be simpler


petemill:
`hasErrors` should not be exposed - FormItem should detect if its slot is filled and check both that and `showErrors` in order to be in an error state. Input and Dropdown, etc should not expose `hasErrors`.

PR #462, opened 2023-10-30
Detect error slot for inputs instead of hasErrors prop

benjaminknox:
Resolves https://github.com/brave/leo/issues/455

Currently to show error messages on inputs, `hasErrors` prop is passed. This uses `$$slots.errors` instead of exposing the `hasErrors` prop, implemented on `Dropdown` and `Input` components.

### With slot present
Dropdown:
Screenshot 2023-10-30 at 11 05 58 AM

Input:
Screenshot 2023-10-30 at 11 05 36 AM

### Without slot present

Dropdown:
image

Input:
image

### DIFF
--- a/src/components/dropdown/dropdown.svelte
+++ b/src/components/dropdown/dropdown.svelte
@@ -102,7 +102,7 @@
     <slot />
   </Menu>
 </div>
-{#if showErrors && $$slots.errors}
+{#if showErrors}
   <slot name="errors" />
 {/if}
 
--- a/src/components/dropdown/dropdown.svelte
+++ b/src/components/dropdown/dropdown.svelte
@@ -29,7 +29,6 @@
   export let required = false
   export let mode: Mode = 'outline'
 
-  export let hasErrors = false
   export let showErrors = false
 
   let dispatch = createEventDispatcher()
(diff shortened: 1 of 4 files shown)


src/components/input/input.svelte
```diff
@@ -164,7 +156,7 @@
   </slot>
   <slot slot="label" />
 </FormItem>
-{#if showErrors && hasErrors}
+{#if showErrors && $$slots.errors}
```
fallaciousreasoning, line 159: I don't think we should need the `$$slots.errors` here - it's like saying, if we don't have the slot, don't show the slot (which won't happen anyway)
```suggestion
{#if showErrors}
```

src/components/dropdown/dropdown.svelte
```diff
@@ -103,7 +102,7 @@
     <slot />
   </Menu>
 </div>
-{#if showErrors && hasErrors}
+{#if showErrors && $$slots.errors}
```
fallaciousreasoning, line 105:
```suggestion
{#if showErrors}
```

I don't think we need the `$$slots.errors` check :smile:

fallaciousreasoning (commented):
This looks great! Couple of minor changes and we should be good!

src/components/input/input.svelte
benjaminknox: Ah, yes thanks for pointing that out, makes sense

fallaciousreasoning (approved):
LGTM!

PR #464, opened 2023-10-31
[Tooltip]: Validates arrow positioning / adds timeout props

benjaminknox:
Right now in main, if you hover over a tooltip and hover out a few times, the arrow positioning sometimes computes incorrectly. This fixes it by validating arrow positioning before firing the `computedposition` event on the `Floating` element.

This PR also adds a few other things discussed with @fallaciousreasoning, a prop named `mouseleaveTimeout` for the debounce timeout on the mouse leave event, and defaults to no fadeout animation on the tooltip, with a prop named `fadeDuration` to set it if desired.

#### Arrow Position Bug Before

https://github.com/brave/leo/assets/5668789/db216dc8-16eb-48cf-bd8e-6256dea8656c

#### Arrow Position Bug After

https://github.com/brave/leo/assets/5668789/9d1c07d5-8694-4436-a6f2-56845b6507ef

https://github.com/brave/leo/assets/5668789/d071c71b-5bbc-4d00-8243-8bd8a425039d

### DIFF
--- a/src/components/floating/floating.svelte
+++ b/src/components/floating/floating.svelte
@@ -60,6 +60,10 @@
     return result
   }
 
+  function hasInvalidArrowPosition({ x, y }: { x?: number; y?: number }) {
+    return (!x && !y) || x <= 1 || y <= 1
+  }
+
   function updatePosition(...args: any[]) {
     if (!floating || !target) return
 
@@ -67,6 +71,8 @@
       placement: placement,
       middleware: getMiddlewares(flip, shift, offset, middleware)
     }).then(({ x, y, placement, middlewareData }) => {
+      if (hasInvalidArrowPosition(middlewareData.arrow)) return
+
       if (floating) {
         Object.assign(floating.style, {
           left: `${x}px`,
--- a/src/components/floating/floating.svelte
+++ b/src/components/floating/floating.svelte
@@ -60,6 +60,9 @@
     return result
   }
 
+  // We need this check because the middleware sometimes generates an
+  // x or y value of undefined, 1, or less. This causes the arrow on
+  // the floating element to be positioned in the corner incorrectly.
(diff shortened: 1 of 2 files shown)


src/components/floating/floating.svelte
```diff
@@ -60,13 +60,19 @@
     return result
   }
 
+  function hasInvalidArrowPosition({ x, y }: { x?: number; y?: number }) {
+    return (!x && !y) || x <= 1 || y <= 1
```
fallaciousreasoning, line 64: Does this work with just
```suggestion
    return !(x || y)
```
or is it important that  we check the `<=` 1?

src/components/tooltip/tooltip.svelte
```diff
@@ -18,9 +18,17 @@
...
+  /** The length of time the tooltip is open in ms after mouse leave of
+   * the trigger or tooltip */
+  export let mouseleaveTimeout: number = 150
+
   /* Whether the tooltip is currently visible */
   export let visible: boolean | undefined = undefined
 
+  /** The length of time in ms the tooltip element takes to fade
+   * after the users mouse leaves the trigger or tooltip*/
+  export let fadeDuration: number = 0
```
fallaciousreasoning, line 30: Nice!

fallaciousreasoning (approved):
LGTM, with one question :smile:

src/components/floating/floating.svelte
benjaminknox: It doesn't actually, that's what I had initially too, the middleware returns a 1 occasionally too, particularly on the y axis.

But with the check for 1 or less I didn't see anymore results that didn't make sense from the middleware

fallaciousreasoning: Good enough for me - maybe add a comment explaining why we need this check and let's merge.

benjaminknox: Done!

PR #472, opened 2023-11-11

brave-support-admin:

- has a raster image element
- has a raster image element
- has a raster image element
Ignoring icon openllama-color.svg because:
- has a raster image element


icons/product-leo-size-36px.svg
fallaciousreasoning, line 1: @aguscruiz these should probably be marked as color icons if we want to maintain the background color

icons/product-leo-size-72px.svg
fallaciousreasoning, line 1: @aguscruiz, these different sized icons all look very similar (I guess the glow is slightly different)

icons/product-search-size-36px.svg
fallaciousreasoning, line 1: @aguscruiz same here - these should have a `-color` suffix

icons/type-browser.svg
fallaciousreasoning, line 1: @aguscruiz `-color`

fallaciousreasoning: Also, this seems like a bit of an odd icon :laughing:

icons/type-initials.svg
fallaciousreasoning, line 1: I dunno if this needs to be part of Nala :laughing:

AlanBreck: cc @aguscruiz

icons/product-leo-size-36px.svg
aguscruiz: Will do

icons/type-initials.svg
aguscruiz: Nah, probably not. We can ignore that. Should I remove them from this page?

icons/type-browser.svg
aguscruiz: Yeah it shouldn't be here

icons/type-initials.svg
AlanBreck: Yeah, that way we don't get it automatically included.

icons/product-leo-size-72px.svg
AlanBreck: @aguscruiz, nudging on ☝️

aguscruiz:
> LGTM!
> 
> Noting that there are multiple removals. Not using any in my projects, but calling out in case their being used any where in the browser. cc @fallaciousreasoning @petemill

I think the ones I removed aren't really used anywhere. Let me know if you found any instance, but I'm pretty sure they're unused

fallaciousreasoning (approved):
LGTM - looks like brave-core has a dependency on `sync` but I don't think it's actually using it :smile:

PR #475, opened 2023-11-14
[Dependencies]: Upgrade to Svelte 4 & Storybook 7

fallaciousreasoning:
Unfortunately, these two dependencies have to be upgraded together, which is a little bit of a hassle.

Additionally:
- Converted Storybook to build with Vite (as obviously, the Storybook Svelte [Webpack builder doesn't support Typescript](https://github.com/storybookjs/addon-svelte-csf/pull/41#issuecomment-1659649429) (advice is to use Vite))
- Converted Storybook default args system


tsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
   "extends": "@tsconfig/svelte/tsconfig.json",
   "include": ["src/components/**/*", "src/svelteDirectives/**/*"],
-  "exclude": ["node_modules/*", "__sapper__/*", "public/*"]
+  "exclude": ["node_modules/*", "__sapper__/*", "public/*"],
+  "compilerOptions": {
+    "strictNullChecks": false,
+    "noImplicitAny": false
```
AlanBreck, line 7: Just curious, why are we disabling `noImplicitAny`?

AlanBreck (approved):
LGTM!

tsconfig.json
fallaciousreasoning: Because the old `svelte-config` allowed implicit any, and the new one doesn't and I didn't want to convert all our code :laughing: 

As a bonus, I have a bit of a controversial opinion - I don't mind implicit any in the first place.

AlanBreck: > Because the old svelte-config allowed implicit any, and the new one doesn't and I didn't want to convert all our code 😆

Makes sense!

PR #478, opened 2023-11-14
[Dialog]: Conditionally render title row

### LINKED ISSUE #355: [Component] Modal should not leave margin for Title when Title slot is absent
I noticed that when not including the title, there's a space left for the title element. They'll be times where we'll want modals without titles. If this is not the case for design, we should make `title` a required element.

Screenshot 2023-08-22 at 10 30 17 AM

fallaciousreasoning:
Fixes #355


nullhook (approved):
lgtm

AlanBreck (approved):

Issue #477, opened 2023-11-14
[Bug] Tooltip content "scrolls" when against an edge


benjaminknox:
# What?

I noticed the tooltip content appears to "scroll" into line when opened against an edge of the browser.

https://github.com/brave/leo/assets/5668789/184df6a2-179c-4703-8893-16b75aa62918

This seems like a bug to me

petemill:
This is either a duplicate or caused by https://github.com/brave/leo/issues/391 depending on how you look at it

PR #488, opened 2023-11-23
[Tooltip]: Fix scroll on screen using width: max-content css prop

benjaminknox:
Closes https://github.com/brave/leo/issues/477

When the content inside the tooltip is without a defined width, such as when it is text content, there is a scroll effect as the `floating-ui` library computes the max content width - only when along an edge of the browser. Adding `width: max-content;` fixes it for when the tooltip content is just text.

### Before

https://github.com/brave/leo/assets/5668789/453d9630-a15b-4d7a-9a38-f91beaaa73ba

### After

https://github.com/brave/leo/assets/5668789/d54e5348-623e-4ca4-b24a-61e1b7beab51

### DIFF
--- a/src/components/tooltip/tooltip.svelte
+++ b/src/components/tooltip/tooltip.svelte
@@ -173,6 +173,7 @@
   }
 
   .leo-tooltip .tooltip {
+    width: max-content;
     background: var(--background);
     color: var(--text);
     box-shadow: var(--shadow);


AlanBreck (commented):
Thanks, @benjaminknox! This is definitely an annoying behavior. I think the problem with `white-space: nowrap` is that sometimes the tooltip has enough comment that we need it to wrap. I think the solution to this problem may be providing a customizable width css var for the tooltip.

benjaminknox:
@AlanBreck Thanks for pointing that out, I think that when a width isn't defined we could still use the `white-space: nowrap;` property and use normal wrapping when a width is defined. That way you can omit the width property if you want to and everything will work as expected.

Here's a PR against this branch with that code: https://github.com/benjaminknox/leo/pull/2

What do you think of this solution?

fallaciousreasoning:
I think it'd be ideal if we could avoid needing to provide an explicit width, if possible. I wonder if rendering the tooltip with `visibility: hidden` for a 100ms or something would be enough to get `floating-ui` to position the tooltip correctly?

Alternatively we could always render the tooltip with `visibility: hidden/visible` and then hopefully the positioning logic would have time to run?

benjaminknox:
@AlanBreck @fallaciousreasoning Thanks for the reviews! I considered your comments and started looking for another way to solve the problem, after reading the floating-ui docs and there's a bit of related info here: https://floating-ui.com/docs/computePosition#initial-layout.

The solution is pretty simple actually, after adding the `width: max-content;` css prop to the `.tooltip` element the placement appears to be working as intended.


https://github.com/brave/leo/assets/5668789/45bb0a3b-aba7-48d2-83d5-9d07f1603292

fallaciousreasoning (approved):
LGTM :laughing: Looks like a pretty nice solution tbh

AlanBreck:
Nice! Thank you @benjaminknox!

PR #489, opened 2023-11-23
[SegmentedControl]: adds SegmentedControl component

AlanBreck:


src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+
+  onMount(() => {
+    pill.addEventListener("transitionstart", () => {
+      segmentedControl.classList.add("transitioning");
+    });
+
+    pill.addEventListener("transitionend", () => {
+      segmentedControl.classList.remove("transitioning");
+    });
+  });
```
AlanBreck, line 115: This is to avoid there being jank from conflicting styles when hovering an element while the background pill moves to the correct position.

src/components/segmentedControl/segmentedControl.stories.svelte
```diff
@@ -0,0 +1,47 @@
...
+  
+    <leo-segmented-item value="item3">
+      <Icon name="check-circle-outline" />
+      Item
+    </leo-segmented-item>
+  </SegmentedControl>
+</Template>
+
+<Story name="Slots" let:args>
+  <SlotInfo
```
fallaciousreasoning, line 39: We should add explanations of the slots here.

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
+<script context="module" lang="ts">
+  declare global {
+    namespace JSX {
+      interface IntrinsicElements {
+        'leo-segmented-item': HTMLAttributes<HTMLElement> & {
+          // Note: This should line up with Reacts key type, but we don't want
+          // to depend on React in this layer, so we just define it manually.
+          key?: string | number | null
```
fallaciousreasoning, line 8: I wonder if we should share this somewhere, as this is basically the third time we're defining this type.

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+
+  export let selected: string | undefined = undefined
+  export let size: Size = 'default'
+
+  let segmentedControl: HTMLDivElement;
+
+  let pillWidth: number;
+  let pillPosition: number;
+
+  $: console.log(selected)
```
fallaciousreasoning: nit: remove

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+  }
+</script>
+
+<script lang="ts">
+  import type { HTMLAttributes } from 'svelte/elements'
+  import { createEventDispatcher, onMount } from 'svelte'
+
+  type Size = 'default' | 'small' | 'tiny'
+
+  export let selected: string | undefined = undefined
```
fallaciousreasoning, line 24: Optional: We could match the native controls and use `value` instead of `selected` here.

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+
+  let segmentedControl: HTMLDivElement;
+
+  let pillWidth: number;
+  let pillPosition: number;
+
+  $: console.log(selected)
+
+  const dispatch = createEventDispatcher<{
+    'select-item': any
```
fallaciousreasoning, line 33: Optional: `change`, to match other events - also, should the detail type be something like `{ value: string | undefined }`

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+
+  function getValue(e: Element) {
+    // If the option element doesn't have a value, fallback to using the text
+    // content - this allows writing simplified options:
+    // i.e. <o>1</o>
+    return e.getAttribute('value') ?? e['value'] ?? e.textContent
+  }
+
+  $: controlItems = Array.from(
+    // TODO(petemil): support slot *descendents* that are leo-menu-item or leo-option (or something else unstyled like leo-menu-action?) so that we can
```
fallaciousreasoning, line 44: nit: I think we can remove this comment from here

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+    // select items that are in complex positions (see an example in the browser app menu zoom controls).
+    (segmentedControl?.querySelector('slot') as HTMLSlotElement)
+      ?.assignedElements()
+      ?.filter((element) =>
+        ['LEO-SEGMENTED-ITEM'].includes(element.tagName)
+      ) ??
+      segmentedControl?.querySelectorAll(
+        ':is(leo-segmented-item)'
```
fallaciousreasoning, line 52: Not super strong on my CSS but is there any advantage to using `is` here, over just querying `leo-segmented-item`?

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+      ) ??
+      []
+  ) as HTMLElement[]
+
+  $: {
+    for (const controlItem of controlItems) {
+      controlItem.setAttribute('tabindex', '0')
+
+      if (controlItem.tagName === 'LEO-SEGMENTED-ITEM') {
```
fallaciousreasoning, line 61: I don't think `controlItem` can ever not be a `LEO-SEGMENTED-ITEM`, can it?

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+        controlItem.setAttribute('role', 'option')
+
+        if (selected === getValue(controlItem)) {
+          controlItem.setAttribute('aria-selected', '')
+          pillWidth = controlItem.clientWidth;
+          pillPosition = controlItem.offsetLeft;
+        } else controlItem.removeAttribute('aria-selected')
+      } else {
+        controlItem.setAttribute('role', 'menuitem')
```
fallaciousreasoning, line 70: I don't think this branch can be hit can it?

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+
+  function selectItem(e: Event) {
+    // Find the option which was clicked on, if any.
+    const item = controlItems.find((item) => e.composedPath().includes(item))
+
+    // If the event was triggered for something which isn't an option don't fire
+    // a change event.
+    if (!item) return
+
+    // Use data-is-interactive=true to prevent the menu from closing when selected. This implies
```
fallaciousreasoning, line 83: I don't think this comment applies, does it? We don't support interactivity inside our segmented control.

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+
+      dispatch('select-item', {
+        value: selected
+      })
+    }
+
+    // When using keyboard navigation to call selectItem, ensure click handlers happen on the item,
+    // but not when this was called in a click handler, otherwise we'll get 2x clicks.
+    if (item.tagName === 'LEO-SEGMENTED-ITEM' && e.type !== 'click') {
+      item.click()
```
fallaciousreasoning, line 99: I'm having a bit of a brain fart, I can't remember why we do this. From memory, selecting something (like a button) with the keyboard actually triggers a click event....

fallaciousreasoning: That's probably from the dropdown, now I'm thinking about it, so this control might be fine without it.

src/components/segmentedControl/segmentedControl.svelte
fallaciousreasoning: I'm still seeing a little bit of jank on my machine (but weirdly only in dark mode). I wonder if wrapping this in a `setTimeout` will help (but I'm clutching at straws).

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+    pill.addEventListener("transitionend", () => {
+      segmentedControl.classList.remove("transitioning");
+    });
+  });
+</script>
+
+<div
+  bind:this={segmentedControl}
+  class="leo-segmented-control"
+  class:isSmall={size === 'small'}
```
fallaciousreasoning, line 119: I'm not a huge fan of this size convention, tbh. Most of our classes are `kebab-case` but for `sizes` and `kinds` we use camel case.

What about if we did `class="leo-segmented-control size-{size}" or something?

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,222 @@
...
+  class:isTiny={size === 'tiny'}
+  on:keypress={(e) => {
+    if (e.code !== 'Enter' && e.code !== 'Space') return
```
fallaciousreasoning, line 122: I have a sneaking suspicion that we don't actually need this, if the items are focusable, but we'll have to test to confirm.

AlanBreck:
I believed I found a fix which can seen in this commit: https://github.com/brave/leo/pull/489/commits/10bf1b17bbbbb7e1e98e26e5aeb361c14970cf20

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,225 @@
...
+    selectItem(e)
+  }}
+  on:click={selectItem}
+>
+  <div class="pill" style:width={`${pillWidth}px`} style:left={`${pillPosition}px`} bind:this={pill} />
```
fallaciousreasoning, line 127: This is a bit of an odd question, but I'm wondering if we could have a way for the consumer to override the highlight. This control feels quite similar to the navdots and I can imagine a scenario where we want to bind the highlight to something like the scroll pos.

Probably not too important (we can merge without) but maybe worth thinking about?

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,225 @@
...
+  <slot />
+</div>
+
+<style lang="scss">
+  .leo-segmented-control {
+    --leo-icon-size: var(--leo-icon-m);
+    --bg: var(--leo-color-container-highlight);
+    --control-padding: var(--leo-spacing-s);
```
fallaciousreasoning, line 135: Probably worth having this configurable - padding is a pretty common tweak.

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,225 @@
...
+      background: var(--leo-color-container-background);
+      position: absolute;
+      height: calc(100% - (var(--control-padding) * 2));
+      min-width: var(--control-height);
+      border-radius: calc(var(--control-height) - var(--control-padding));
+      transition: width 0.2s cubic-bezier(0.22, 1, 0.36, 1), left 0.4s cubic-bezier(0.22, 1, 0.36, 1);
+    }
+
+    :global :where(&) ::slotted(leo-segmented-item),
+    :global :where(&) > leo-segmented-item {
```
fallaciousreasoning, line 175: Just confirming that you've checked everything is all good in WebComponent land?

fallaciousreasoning (commented):
Looks great!

src/components/segmentedControl/segmentedControl.svelte
AlanBreck: I guess the question would be _should_ it get overridden. @aguscruiz?

src/components/segmentedControl/segmentedControl.svelte
AlanBreck: Should it be tweaked, though?

src/components/controlItem/controlItem.svelte
```diff
@@ -0,0 +1,60 @@
...
+    color: var(--leo-control-item-color);
+    font: var(--leo-control-item-font);
+    background: var(--leo-control-item-background);
+    box-shadow: var(--leo-control-item-shadow);
+    position: relative;
+    z-index: 10;
+
+    &,
+    & :global(.icon),
+    :global & ::slotted(.icon) {
```
AlanBreck, line 40: I cannot manage to get this to work with web components.

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -0,0 +1,182 @@
...
+  ) as HTMLElement[]
+
+  $: {
+    for (const controlItem of controlItems) {
+      controlItem.setAttribute('role', 'option')
+
+      if (value === getValue(controlItem)) {
+        controlItem.setAttribute('aria-selected', '')
+        pillWidth = controlItem.getBoundingClientRect().width
+        pillPosition = controlItem.offsetLeft
```
AlanBreck, line 44: I don't understand why, but for some reason, the initial load grabs the incorrect width and offset when using web components.

AlanBreck: @petemill recommends messing with request animation frame.

petemill:
Just to summarize what we just discussed: The advantage of using Components for children is that the styling is simplified - the children can decide how to be styled for regular, hover, focus, selected, etc states, and we don't have to resort to complex selectors which are required to use custom elements styled directly from the parent / sole component. So, I'm not sure if there's much point to having child Components but still keep the styling in the parent.

I get that the issue is that we can't tell the child it's selected from the parent when the child is a Component, because we don't have Context that works in both WC and Svelte land yet....

So I think you can decide whether to keep child Components or custom elements, whatever is easier for you to get working in WC-land and Svelte-land! 😄

src/components/segmentedControl/segmentedControl.svelte
AlanBreck: That worked!

AlanBreck:
Fixed those web component issues.

I opted to stick with the component since it's not necessarily worse than the custom elements, and is closer to what we'll ultimately want.

src/components/segmentedControl/segmentedControl.svelte
fallaciousreasoning: The reality with the `Button` has been that it will be (at least a little bit) in different designs. I wouldn't be surprised if it's pretty similar for this

src/components/controlItem/controlItem.svelte
fallaciousreasoning, line 1: Mega nit - but `controlItem` sounds like it can do more than it actually can. This can only go inside the segmented control.

I'd be tempted to:
1. Move this into the `segmentedControl` folder
2. Rename it to (the sadly verbose) `SegmentedControlItem`

This should make it a bit more obvious that they belong together.

src/components/controlItem/controlItem.svelte
```diff
@@ -0,0 +1,60 @@
+<script lang="ts">
+  export let value: any
```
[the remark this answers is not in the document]
fallaciousreasoning: And `number` is tenuous, as it's going to be dumped in the dom...

src/components/segmentedControl/segmentedControl.stories.svelte
```diff
@@ -1,46 +1,61 @@
...
-  let selected = "item1";
+  let selected = 'full'
 </script>
 
 <Meta
   title="Components/SegmentedControl"
   component={SegmentedControl}
-  argTypes={{}}
+  argTypes={{
+    size: {
```
fallaciousreasoning, line 19: Nit: Add CSS properties here.

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -80,63 +55,60 @@
...
+    style:width={`${pillWidth}px`}
+    style:left={`${pillPosition}px`}
+    bind:this={pill}
+  />
   <slot />
 </div>
 
 <style lang="scss">
+  :host {
+    position: relative;
```
fallaciousreasoning, line 98: Is `inline-block` an appropriate display for this component (unfortunately all web components default to inline-block)?

fallaciousreasoning (approved):
LGTM with some nits. I guess we should get some kind of context working in WebComponents land because some of this stuff is a bit painful.

src/components/controlItem/controlItem.svelte
AlanBreck: The idea is that this component would also be used for the other controls here: https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?type=design&node-id=278-12661&mode=dev

src/components/controlItem/controlItem.svelte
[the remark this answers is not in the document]
AlanBreck: In WC land, yes. In Svelte land, it can be anything. You're probably right, though, that string/number is more appropriate.

src/components/segmentedControl/segmentedControl.svelte
AlanBreck: What? Oof. No, it should just be `block`.

src/components/controlItem/controlItem.svelte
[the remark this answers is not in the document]
fallaciousreasoning: I don't think it can be because we're storing it in the DOM, so if it's something more complex (like an array or object) it'll get converted to a (not necessarily useful) string, like `[Object]`

src/components/segmentedControl/segmentedControl.svelte
fallaciousreasoning: I've shot myself with this quite a few times :D

src/components/controlItem/controlItem.svelte
fallaciousreasoning: Hmm okay, I guess fine to leave it on it's own :laughing: Wish we had a better name than `ControlItem` though - it sounds like it should be used with `Input` or `TextArea` or something

AlanBreck: >  it sounds like it should be used with Input or TextArea or something

Why's that?

fallaciousreasoning: I think of those things as `Controls` - basically something that goes in a form :laughing: I guess my point is just that `ControlItem` sounds like it can be used with any control (range, input, color, password, date), but it can only be used with a few specific ones (segmentedButtons, buttonGroups, tabs)

AlanBreck: We could considering solving via folder structure:

```
components/
└── segmentedControls/
    ├── controlItem/
    │   └── controlItem.svelte
    ├── segmentedControl/
    │   └── segmentedControl.svelte
    ├── buttonGroup/
    │   └── buttonGroup.svelte
    └── tabGroup/
        └── tabGroup.svelte
```

fallaciousreasoning: I think when everything is compiled the structure is completely flat :laughing: I think what you have will be fine for now :smile:

AlanBreck: Right. If we ultimately wanted something like that, we'd have to change the compiled structure as well.

src/components/segmentedControl/segmentedControl.svelte
AlanBreck: It's necessary since the parent is managing the click event.

src/components/segmentedControl/segmentedControl.svelte
fallaciousreasoning: Ooooh flex - Mr Fancy :laughing:

fallaciousreasoning (approved):
LGTM, thanks @AlanBreck!

Issue #484, opened 2023-11-21
Update color rules for buttons and hover states


aguscruiz:
### New color rules

Based on requirements from the design team, we'll need to be able to let the colors of the button change depending on different needs. Colors should be customizable to reflect different states

![image](https://github.com/brave/leo/assets/15734630/ae539ee7-aa1d-48da-b3b1-c20459552c1b)

### New hover states

This new set of customizations puts us into some trouble when it comes to hover states. @fallaciousreasoning were talking about using [-color mix](https://www.smashingmagazine.com/2022/06/simplify-color-palette-css-color-mix/) as a solution for the issue, with the following values:

Light mode: Add a mix of 20% black to the colors.
Dark mode: Add a mix of 20% white to the colors

<img width="1324" alt="image" src="https://github.com/brave/leo/assets/15734630/4ff5854d-cb64-462d-9eb4-9823de466796">

<img width="1338" alt="image" src="https://github.com/brave/leo/assets/15734630/781f8e6a-e9c8-4257-b8c7-c2e2b9ec8662">

fallaciousreasoning:
Cool this should be quite an interesting change - I don't know if we can use `color mix` yet, as all browsers only added support in May 2023. @petemill wdyt? Alternatively, we could generate a palette from the color at build/run time?

AlanBreck:
We could potentially do something similar, but use `hsl` instead of `color-mix`. The below is a little clunky, but we could maybe create a PostCSS plugin for a convenience function.

https://codepen.io/AlanBreck/pen/poGVEoL

fallaciousreasoning:
@aguscruiz how do we handle `:focus` `:active`, `disabled` and `loading` states?

aguscruiz:
Those states can remain they are now, I think

PR #490, opened 2023-11-27
[Button]: Experimental support for Button palettes

fallaciousreasoning:
Draftish still. Wondering what you guys think of the approach - it seems to work pretty nicely, and `color-mix` has support in all browsers since [May 2023](https://caniuse.com/?search=color%20mix)

Fixes https://github.com/brave/leo/issues/484

Settings `--leo-button-color` will generate a palette for
- `filled`
- `outline`
- `plain`
Buttons. `hero` and `plain-faint` buttons are not themable. If the browser doesn't support `color-mix` then `--leo-color-button-background` is used as the theme.


src/components/button/button.svelte
```diff
@@ -145,6 +159,17 @@
...
     [data-is-button-target]:hover .leoButton {
       --leo-icon-color: var(--icon-hover-color, var(--icon-color));
+      --mixed-primary-color: var(--leo-color-primitive-primary-70);
+
+      /** If we support the color-mix syntax, infer the hover color */
+      @supports (color: color-mix(in srgb, transparent, transparent)) {
+        --mixed-primary-color: color-mix(
+          in srgb,
+          var(--primary-color) 80%,
+          var(--foreground)
```
AlanBreck, line 169: Would it make more sense here to do the following since the "spec" is 20% black or white?

```suggestion
          var(--primary-color),
          var(--foreground) 20%
```

AlanBreck (commented):
This is looking really cool!

fallaciousreasoning:
@petemill @aguscruiz has said he's a fan of this - do we want to take it further?

petemill (approved):
Pretty cool! We just discussed it and we'll have to keep an eye on how it goes. We should communicate this new property to people that might encounter different colors in designs.

src/components/button/button.svelte
```diff
@@ -108,6 +108,18 @@
   :host button {
     width: 100%;
   }
+
+  .leoButton {
+    --foreground: var(--leo-color-text-primary);
+    --background: var(--leo-color-container-background);
```
petemill, line 114: WE were just discussing how these variable names are pretty finite in that they seem like they will be used for `foreground` and `background` on the button, but they are simply the foreground and background "base" colors that will be mixed with other colors. Perhaps at least just a comment here saying that's what they are, as I can't really think of a better variable name - they are understandable after all.

PR #493, opened 2023-11-29
misc fixes to various components

AlanBreck:
@petemill and @fallaciousreasoning, these are miscellaneous fixes that have been piling up as I'm working on the accounts site. I put them all in one commit in this one PR just to get the ball rolling, but what's the best way to split this out? One PR per component?

**Update:** I've split out the work into distinct commits. I'd recommend that when we merge, we rebase instead of squash.


src/components/menu/menu.svelte
```diff
@@ -254,23 +256,29 @@
...
     background: var(--leo-color-container-interactive);
     color: var(--leo-color-text-interactive);
   }
 
-  :global .leo-menu-popup ::slotted(*:focus-visible),
-  :global .leo-menu-popup > *:focus-visible {
+  :global :where(.leo-menu-popup) ::slotted(leo-option:focus-visible),
+  :global :where(.leo-menu-popup) ::slotted(leo-menu-item:focus-visible),
+  :global :where(.leo-menu-popup) > leo-option:focus-visible,
+  :global :where(.leo-menu-popup) > leo-menu-item:focus-visible {
```
AlanBreck, line 281: Though more verbose, this makes it easier to have other content in the menu without having to reset their color/background color.

src/components/menu/menu.svelte
```diff
@@ -254,23 +256,29 @@
...
   }
 
-  :global .leo-menu-popup ::slotted(leo-menu-item:hover),
-  :global .leo-menu-popup ::slotted(leo-option:hover),
-  :global .leo-menu-popup > leo-menu-item:hover,
-  :global .leo-menu-popup > leo-option:hover {
+  :global :where(.leo-menu-popup) ::slotted(leo-menu-item:hover),
+  :global :where(.leo-menu-popup) ::slotted(leo-option:hover),
+  :global :where(.leo-menu-popup) > leo-menu-item:hover,
+  :global :where(.leo-menu-popup) > leo-option:hover {
```
AlanBreck, line 262: `:where()` reduces the specificity, which makes it easier for the consumer override styles.

src/components/dropdown/dropdown.svelte
```diff
@@ -48,11 +48,12 @@
 </script>
 
 <div class="leo-dropdown">
-  <div bind:this={dropdown}>
+  <div>
     <FormItem
       bind:disabled
       bind:required
       bind:size
+      bind:controlElement={dropdown}
```
AlanBreck, line 56: Previously, the popup would expand to the width of the control _and_ the label, if the label was horizontally aligned. This ensures that the popup width is based on the control, and not the label.

src/components/button/button.svelte
```diff
@@ -294,15 +294,25 @@
...
     --color-hover: var(--leo-color-gray-70);
     --box-shadow-hover: none;
-    --icon-color: var(--leo-color-icon-default);
+    --icon-color: var(--leo-button-faint-color, var(--leo-color-icon-default));
     --icon-hover-color: var(--leo-color-gray-70);
+
+    &.fab {
+      --padding-y: 0;
+      --padding-x: 0;
+    }
```
AlanBreck, line 315: The padding for icon only, plain faint buttons is different than the others, according to Figma.

src/components/button/button.svelte
```diff
@@ -294,15 +294,25 @@
...
   }
   .leoButton.isPlainFaint {
     --radius: 8px;
     --padding-x: 2px;
-    --color: var(--leo-color-text-secondary);
+    --color: var(--leo-button-faint-color, var(--leo-color-text-secondary));
     --color-hover: var(--leo-color-gray-70);
     --box-shadow-hover: none;
-    --icon-color: var(--leo-color-icon-default);
+    --icon-color: var(--leo-button-faint-color, var(--leo-color-icon-default));
```
AlanBreck, line 309: The apparent intent of the plain-faint button is that it _can_ use the same color as the surrounding text. This adds a variable to set that color.

AlanBreck: Discussed that we should wait to add this until https://github.com/brave/leo/pull/490 lands.

src/components/alert/alertCenter.svelte
```diff
@@ -4,16 +4,25 @@
...
+  } & (
+    | {
+        text: string
+        fabIcon?: never
+      }
+    | {
+        fabIcon: IconName
+        text?: never
+      }
+  )
```
AlanBreck, line 17: Allow both `text` and `icon`, and if only `icon` is provided, use `fab` styles.

src/components/dialog/dialog.svelte
AlanBreck, line 1: Remove padding on parent `dialog` and change background to `transparent`.

src/components/menu/menu.svelte
```diff
@@ -279,8 +287,6 @@
     .leo-menu-popup
     > ::slotted(*:hover:not(leo-menu-item):not(leo-option)),
   :global .leo-menu-popup > *:hover:not(leo-menu-item):not(leo-option) {
-    color: inherit;
-    background: none;
```
AlanBreck, line 283: Remove entire rule, and remove `cursor: pointer` from `.leo-menu` rule.

src/components/alert/alertCenter.svelte
```diff
@@ -4,16 +4,25 @@
 
   type Action = {
     kind?: ButtonKind
-    text: string
     action: (alert: AlertInfo) => void
-  }
-
-  interface AlertOptions {
+  } & (
+    | {
```
fallaciousreasoning, line 9:
```suggestion
    {
```

surprised this syntax is valid :laughing:

src/components/formItem/formItem.svelte
```diff
@@ -71,7 +79,7 @@
     <slot name="label" />
     {#if required}<span class="required-indicator">*</span>{/if}
   </div>
-  <div class="control">
+  <div class="control" bind:this={_controlElement}>
```
fallaciousreasoning, line 82: Out of interest, does it work if we just bind directly to `controlElement`? I think if you change it in the parent, you're asking for trouble, so I'm not too concerned if you get it :laughing:

src/components/icon/icon.svelte
```diff
@@ -22,7 +22,7 @@
     // 404).
     if (isSSR) return
 
-    requestIdleCallback(() => {
+    window.requestIdleCallback?.(() => {
```
fallaciousreasoning, line 25: Boooo Safari!

src/components/icon/icon.svelte
```diff
@@ -56,7 +56,7 @@
 <style lang="scss">
   .leoIcon {
     --icon-width: var(--leo-icon-size, 24px);
-    --icon-height: var(--leo-icon-size, 24px);
+    --icon-height: var(--leo-icon-height, var(--leo-icon-size, 24px));
```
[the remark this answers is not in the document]
fallaciousreasoning: Maybe we should do that for `width` too? (i.e. you can specify both with `--leo-icon-size` but override one or the other with width/height?

src/components/alert/alertCenter.svelte
AlanBreck: Yeah! I found it weird too, but that's what `prettier` did.

src/components/alert/alertCenter.svelte
```diff
@@ -83,20 +96,30 @@
...
               kind={action.kind || 'filled'}
-              on:click={() => action.action(alert)}>{action.text}</Button
+              on:click={() => action.action(alert)}
             >
+              {#if action.icon && !action.text}
+                <Icon name={action.icon} />
+              {:else}
+                {action.text}
+              {/if}
+              <div slot="icon-after" class:hide={!action.text}>
```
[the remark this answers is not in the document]
AlanBreck: Sadly, no. `slot="x"` can't go inside of a condition.

src/components/formItem/formItem.svelte
AlanBreck: Binding directly to `controlElement` would definitely work. I just preferred not to give the consumer an opportunity to mess it up.

src/components/alert/alertCenter.svelte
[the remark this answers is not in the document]
fallaciousreasoning: Ah Svelte - why are you like this?

src/components/formItem/formItem.svelte
fallaciousreasoning: Either way is fine - I don't think we can stop people shooting themselves in the foot if they really want, and I don't know what else you'd expect to happen if you change the type of an element (maybe nothing...)

fallaciousreasoning: Lol, you can't even change the element if you want:
https://svelte.dev/repl/c9c159a994444f62a661a33dbdc24709?version=4.2.8

AlanBreck: Interesting! Seems to prevent you from completely replacing the element, but you can still manipulate the DOM element itself. Do you see a particular reason why _not_ to do the above, though?

src/components/alert/alertCenter.svelte
AlanBreck: For whatever reason, my local config for prettier is doing that. Not sure why it's not respecting the repo config. 🤔

src/components/formItem/formItem.svelte
fallaciousreasoning: I don't think it buys us anything, just makes it slightly harder to follow:
1. You already can't replace the element
2. Even with the indirection, you can still modify the DOM element (as you have a reference to it).

AlanBreck: Fair point.

fallaciousreasoning: Either way, feel free to merge!

fallaciousreasoning (approved):
LGTM still @AlanBreck!

fallaciousreasoning (approved):
Still still lgtm :laughing:

PR #496, opened 2023-12-05
[Tokens]: Update Elevations, Font Weights & Colors

aguscruiz:
Updated light/dark modes so they work better.

Icon and text variables are now the same so the icon doesn't stand out much more


fallaciousreasoning:
Note: I think this set of changes will be much less scary once we rebase onto latest `main`

AlanBreck:
Those elevations look really prominent!

aguscruiz:
> Those elevations look really prominent!

I couldn't find a place to see them all in storybook, but they looked ok on the buttons for example, I think

They should look like this
image

AlanBreck (approved):

PR #497, opened 2023-12-06
[Button]: changes plain-faint button to use currentColor

AlanBreck:


src/components/button/button.svelte
```diff
@@ -326,11 +326,18 @@
   .leoButton.isPlainFaint {
     --radius: 8px;
     --padding-x: 2px;
-    --color: var(--leo-color-text-secondary);
-    --color-hover: var(--leo-color-gray-70);
+    --primary-color: currentColor;
```
fallaciousreasoning, line 329: I think the hover color might look a bit funky when we don't support `color-mix` - I think we should maybe override `--mixed-primary-color` (as it'll currently default to blurple)

Ideally, this will look more or less the same, unless we support `color-mix`

fallaciousreasoning: But maybe it doesn't need to look exactly the same - kinda hard to decide what to do with `currentColor` on hover :laughing:

AlanBreck: Based on the Figma for `plain-faint`, the distinction on hover is already super subtle, so I'm wondering if it's alright: https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?type=design&node-id=8010-16891&mode=dev

fallaciousreasoning: cc @aguscruiz

AlanBreck: Examples of the proposed `plain-faint` behavior can be found here: https://497.pr.nala.bravesoftware.com/?path=/story/components-alert--all

AlanBreck: To clarify, @aguscruiz, are you saying that the way it's behaving in Storybook is how you'd want it, and therefore we could theoretically proceed with this PR as is?

fallaciousreasoning:
@aguscruiz so are you saying we should merge this `plain-faint` variant with no hover state @aguscruiz? 

@AlanBreck I kind of feel like we should try and add one, at least when color mix is supported. Currently we use these buttons a lot in Brave-Core, and I think it's a bit of a regression losing out on the visual affordance.

Happy to defer to @aguscruiz here though.

src/components/button/button.svelte
AlanBreck: We decided in our call to mix `transparent` in for `plain-faint` buttons.

fallaciousreasoning (approved):

PR #503, opened 2023-12-07
[Storybook]: removes unused fonts from storybook config

AlanBreck:


fallaciousreasoning (commented):
Poppins is still used for headings, right, just not those weights?

fallaciousreasoning (approved):

PR #504, opened 2023-12-07
[Floating]: improves flexibility of floating component

AlanBreck:


src/components/menu/menu.svelte
```diff
@@ -222,9 +222,7 @@
 
     border: 1px solid var(--leo-color-divider-subtle);
     border-radius: var(--leo-radius-m);
-    width: max-content;
     min-width: var(--leo-menu-control-width);
-    max-width: var(--leo-menu-popup-max-width);
```
fallaciousreasoning, line 227: Again, this feels slightly off to me. Is there a reason we can't control the width of the slotted content?

src/components/tooltip/tooltip.svelte
```diff
@@ -173,7 +177,6 @@
   }
 
   .leo-tooltip .tooltip {
-    width: max-content;
```
fallaciousreasoning, line 176: I have a feeling that removing this is going to reintroduce the layout shift which was fixed by https://github.com/brave/leo/pull/488

src/components/floating/floating.svelte
```diff
@@ -110,14 +119,14 @@
...
   <slot />
 </div>
 
 <style lang="scss">
   .leo-floating {
-    position: absolute;
     z-index: 999;
-    width: fit-content;
+    width: max-content;
+    max-width: var(--leo-floating-max-width);
```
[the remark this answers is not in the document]
AlanBreck: I'd have to test to be sure, but I think that'd be fine. Seems a reasonable knob for this floating component concept, and I generally don't prefer scenarios where I'm forced to add an extra div just to do something like this.

Again, though, happy to change if preferred.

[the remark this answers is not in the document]
fallaciousreasoning: Let's get @petemill's opinion on this, 'cause it might just be me :laughing:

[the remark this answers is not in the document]
AlanBreck: We decided in our call to strike this for now.

src/components/buttonMenu/buttonMenu.stories.svelte
```diff
@@ -12,7 +12,15 @@
   const handleAction = () => console.log('action')
 </script>
 
-<Meta title="Components/ButtonMenu" component={ButtonMenu} />
+<Meta
+  title="Components/ButtonMenu"
+  component={ButtonMenu}
+  argTypes={{
+    '--leo-menu-control-width': {
```
fallaciousreasoning, line 19: I don't think this exists anymore

fallaciousreasoning (approved):

src/components/buttonMenu/buttonMenu.stories.svelte
AlanBreck: This one still exists. It's a read only prop to know the width of the control. https://github.com/brave/leo/blob/ecc35e8b9f5f8cce355d7b8da0bec86ee3cf7ebf/src/components/menu/menu.svelte#L184

Issue #511, opened 2023-12-19
Updates to segmented control component


aguscruiz:
- [ ] Changed the rounded corner values for the internal element and external container.
- [ ] Changed padding for both elements and sizing.
- [ ] Added elevation/01 to active state

https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?type=design&node-id=7719%3A15980&mode=design&t=VHhtLpJltP9ac2IA-1

Issue #510, opened 2023-12-19
Update button shapes and paddings


aguscruiz:
Updated the shape of buttons in all sizes.

- [ ] Different button sizes now have different rounded corner values.
- [ ] Paddings were updated for all the sizes.
- [ ] This affects regular buttons and icon only buttons.

https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?type=design&node-id=278%3A12159&mode=design&t=VHhtLpJltP9ac2IA-1

PR #523, opened 2024-01-10
Updates components to rounded rectangle style

### LINKED ISSUE #514: Dropdown updates
- [ ] Updated rounded corner values for each size to match button values
- [ ] Updated heights of the actual inputs to match button values

https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?type=design&node-id=195%3A7599&mode=design&t=VHhtLpJltP9ac2IA-1

### LINKED ISSUE #567: Text field updates
- [ ] Updated rounded corner values for each size to match button values
- [ ] Updated heights of the actual inputs to match button values

https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?type=design&node-id=195%3A7599&mode=design&t=VHhtLpJltP9ac2IA-1

AlanBreck:
Resolves https://github.com/brave/leo/issues/510
Resolves https://github.com/brave/leo/issues/511
Resolves https://github.com/brave/leo/issues/528
Resolves #567
Resolves #514

NOTE: this should be rebased and not squashed.

### DIFF
--- a/src/components/button/button.svelte
+++ b/src/components/button/button.svelte
@@ -136,6 +136,7 @@
     justify-content: center;
     align-items: center;
     cursor: pointer;
+    box-sizing: border-box;
     -webkit-tap-highlight-color: transparent;
     transition:
       background 0.12s ease-in-out,
--- a/src/components/button/button.svelte
+++ b/src/components/button/button.svelte
@@ -74,17 +74,17 @@
   {...$$restProps}
 >
   {#if isLoading}
-    {#if $$slots.loading}
-      <slot name="loading" />
-    {:else if !fab}
-      <div>
+    <div class:content={!fab}>
+      {#if $$slots.loading}
+        <slot name="loading" />
+      {:else if !fab}
         <slot>Leo Button</slot>
-      </div>
-    {/if}
+      {/if}
+    </div>
     <ProgressRing />
   {:else}
     <slot name="icon-before" />
-    <div>
+    <div class:content={!fab}>
       <slot>Leo Button</slot>
     </div>
     <slot name="icon-after" />
@@ -133,7 +133,6 @@
     --leo-progressring-color: var(--icon-color);
 
     display: flex;
-    gap: var(--icon-gap);
     justify-content: center;
     align-items: center;
     cursor: pointer;
@@ -152,6 +151,10 @@
 
     &.fab {
       max-width: max-content;
+    }
+
+    .content {
+      padding: 0 var(--icon-gap);
     }
   }
 
--- a/src/components/button/button.svelte
+++ b/src/components/button/button.svelte
@@ -328,13 +328,22 @@
     --icon-color: var(--color);
   }
   .leoButton.isPlain {
-    --padding-x: 2px;
     --primary-color: var(--leo-button-color, var(--leo-color-primitive-primary-60));
     --color: var(--mixed-primary-color);
     --box-shadow-hover: none;
+    --bg-hover-mix: 5%;
 
     @theme (dark) {
       --primary-color: var(--leo-button-color, var(--leo-color-primitive-primary-40));
+      --bg-hover-mix: 10%;
+    }
+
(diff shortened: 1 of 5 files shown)


src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -39,10 +39,16 @@
...
+        function setPill(t: number) {
+          const timeDiff = t - zero;
           pillWidth = controlItem.getBoundingClientRect().width
           pillPosition = controlItem.offsetLeft
-        })
+          if (timeDiff < 1000) {
+            window.requestAnimationFrame(setPill)
+          }
+        }
+        window.requestAnimationFrame(setPill)
```
AlanBreck, line 51: This may not be a great way to do this, but I'm running into a problem where the initial width isn't _quite_ large enough, so the theory here is to repeatedly request an animation for reasonably long enough period to assume the final width is correct.

aguscruiz (commented):
So cool to see this, couple comments:

- [x] The hero button has an old gradient on hover, instead of darkening the current gradient with the calculation we had set up.
- [ ] About the border radius, I'm gonna leave a link here for discussing later. What I did on the buttons is this weird calculation to smooth the borders (in Figma) but that doesn't come across in straight css it seems. Let me know later, maybe in tomorrow's call, if this is something we can do https://css-houdini.rocks/smooth-corners/ (Search squircle in css for more). 

It's a silly thing to notice, but I promise it makes all the difference on how sleek they look

The rest looks great

aguscruiz (commented):
Looks like the Jumbo buttons are much bigger than what's intended.
In Figma they're 60x60. 

image

-------------------

Same is happening with the rest of the sizes, they seem to be way bigger

AlanBreck:
@aguscruiz, thanks for catching! This should now be resolved. Note that with the incorrect padding removed on icon-only buttons, but hover effect no longer looked good on `plain`/`plain-faint` buttons, so I removed.

aguscruiz (commented):
Sorry, one more comment I didn't catch before. I think the non-plain buttons have their transitions to hover animated, and the plain variants don't 🥺

AlanBreck:
> Sorry, one more comment I didn't catch before. I think the non-plain buttons have their transitions to hover animated, and the plain variants don't 🥺

`plain` buttons are actually currently transitioning the same as non-plain, but `plain-faint` is not. Working on a fix.

src/components/button/button.svelte
```diff
@@ -321,43 +336,85 @@
...
+          var(--primary-color) var(--bg-hover-mix),
+          var(--foreground)
+        );
+        transition: var(--default-transition);
+        opacity: 0.01;
+      }
+
+      &:hover::before {
+        opacity: 1;
+      }
```
AlanBreck, line 412: Note that since `plain-faint` relies on `currentColor`, transitions are working, so I'm using a pseudo element instead in order to transition the opacity of the element instead of the background color.

AlanBreck:
Button size when an `<a>` is really tall.

Sizes of inputs and buttons do not match.

fallaciousreasoning:
@petemill @aguscruiz we should decide how we want to land this in Brave Core. Do we need to update all the native buttons there before this lands? Otherwise it's going to be a bit odd having square buttons some places and round ones in others

aguscruiz:
Ideally, if we can somehow change all the native buttons to squircle too, that'd be a huge win. But there are also other changes to be made, like the overall use of orange being replaced with blurple. 

But, we can do it incrementally. let me know what you think will be better

AlanBreck:
@fallaciousreasoning, by updating the `formItem` component for the `Input`, did that also update the `Dropdown`s, and therefore resolve https://github.com/brave/leo/issues/514?

fallaciousreasoning:
Should have, yeah :smile:

fallaciousreasoning (approved):

Issue #515, opened 2023-12-19
Alerts and toasts - New buttons placement update


aguscruiz:
- [ ] Updated the alerts and toasts so they all have the same button placement now. Below the content itself. Not gonna be besides the content anymore 

https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?type=design&node-id=195%3A5635&mode=design&t=VHhtLpJltP9ac2IA-1

PR #524, opened 2024-01-10
[Alert]: moves actions to bottom of alert regardless of mode

AlanBreck:
Resolves https://github.com/brave/leo/issues/515


aguscruiz (commented):
Great! Looks good to me

fallaciousreasoning:
Are we sure this is what we want? It looks a bit odd for toast notifications
image

Also seems like the space is being reserved even if there are no buttons:
image

AlanBreck:
> Also seems like the space is being reserved even if there are no buttons:

I believe this is only a problem for the `AlertCenter` and shouldn't be hard to address.

src/components/alert/alert.svelte
```diff
@@ -19,6 +19,7 @@
   export let type: AlertType = 'error'
   export let mode: AlertMode = 'simple'
   export let isToast = false
+  export let hasActions = $$slots.actions
```
AlanBreck, line 22: Can't say I'm proud of this, but it's the only way I can see to avoid the empty space when no actions are present for the `<AlertCenter>` given that we can't wrap the `slot="actions"` in a conditional: https://github.com/brave/leo/blob/d296d0335219c8daa942829a28033fdcbd0552b7/src/components/alert/alertCenter.svelte#L101

src/components/alert/alert.svelte
```diff
@@ -68,22 +77,39 @@
...
     color: var(--default-text-color, var(--leo-color-text-primary));
     padding: var(--leo-spacing-xl);
     border-radius: var(--leo-radius-m);
     gap: var(--leo-spacing-xl);
     font: var(--leo-font-default-regular);
 
+    display: grid;
+    grid-template-columns: min-content 1fr;
+
+    &:has(.content-after) {
```
AlanBreck, line 89: How comfortable are we with `:has()`? Looks like most concerning would be Firefox for Android not supporting it yet.

https://caniuse.com/?search=%3Ahas

AlanBreck:
One problem that this surfaces, but does not introduce, is the super wide button in actions due to: https://github.com/brave/leo/blob/8399ec8d09bba37b84b4bceb43119d1b00d8eae5/src/components/button/button.svelte#L108-L110

Screenshot 2024-01-12 at 5 41 29 PM

Not sure of the best solution here.

src/components/alert/alert.svelte
fallaciousreasoning: I think, instead of doing this you could wrap the `<div slot="actions">` in `AlertCenter` in an `#if actions.length` block

fallaciousreasoning: Wait, why can't we wrap that in a conditional?

src/components/alert/alert.svelte
fallaciousreasoning: I think we're fairly comfortable with it, but @petemill to confirm.

src/components/alert/alert.svelte
AlanBreck: It's a big bummer, but Svelte doesn't allow it. Allegedly, `snippets` will solve this problem.

https://github.com/sveltejs/svelte/issues/5604#issuecomment-1808036921

Screenshot 2024-01-16 at 10 50 22 AM

fallaciousreasoning: Wow.... sometimes I really wonder if we should be using Svelte :laughing:

PR #543, opened 2024-02-13
Tokens and variables update 

aguscruiz:
Made a bunch of updates:
- Added a new component font style for large numbers
- Added a hard shadow effect (the actual shadow color is customizable)
- Tweaked the primary color values
- Added RORANGE colors (orange + red)
- Changed the hero gradient

Note for @fallaciousreasoning you can find and replace all Display/3 with Numbers/Large


petemill:
The redirection of cyrillic character sets to Inter should also apply to this new font (via @aguscruiz)

AlanBreck:
Potential blocker for this PR:
- [ ] Use of Flecha (uncertain if possible for browser)
- [ ] Number specific fonts
- [ ] Presence of `browser.variables.json`

AlanBreck:
```diff
-tokens/css/variables-universal.css: 55340 bytes
+tokens/css/variables-browser.css: 3407 bytes
```

fallaciousreasoning (dismissed):

fallaciousreasoning:
Storybook seems broken:
- https://543.pr.nala.bravesoftware.com/?path=/docs/tokens-browser-fonts--docs (fonts have an error)

AlanBreck:
@aguscruiz, just realizing that these versions use Flecha in the universal tokens.

fallaciousreasoning:
> Did we decide to ask that one of these shadows be removed?

I don't really remember. Lets leave it for now, or completely remove the token until we know how to deal with it.

> Also, did we decide we wanted the border baked into the shadow, or up to the consumer?

Let's leave it up the consumer. I think let's keep things as simple as possible for now.

> @aguscruiz, just realizing that these versions use Flecha in the universal tokens.

Okay, we better pull that out before we land.

aguscruiz:
> > Did we decide to ask that one of these shadows be removed?
> 
> I don't really remember. Lets leave it for now, or completely remove the token until we know how to deal with it.
> 
> > Also, did we decide we wanted the border baked into the shadow, or up to the consumer?
> 
> Let's leave it up the consumer. I think let's keep things as simple as possible for now.
> 
> > @aguscruiz, just realizing that these versions use Flecha in the universal tokens.
> 
> Okay, we better pull that out before we land.

@fallaciousreasoning @AlanBreck I was gonna remove it myself and commit but it seems you already did it right? I pulled in the latest and Flecha is gone

src/tokens/transformation/common/leoParser.js
```diff
@@ -2,14 +2,18 @@
...
+const universalVariables = require('../../universal.variables.json')
 const { TinyColor } = require('@ctrl/tinycolor')
 
 module.exports = {
   pattern: /\.json$/,
   parse: ({ filePath, contents }) => {
+    let layerVariables = {}
+    try {
+      layerVariables = require(`${filePath.split('.')[0]}.variables.json`)
+    } catch (_) {}
```
fallaciousreasoning, line 15:
```suggestion
    } catch {}
```

src/tokens/transformation/common/leoParser.js
```diff
@@ -30,12 +34,25 @@ module.exports = {
...
     // At the moment, we just whitelist a few color groups.
     const allowedGroups = ['elevation', 'primary', 'secondary']
-    const effectColors = allowedGroups
+    const universalEffectColors = allowedGroups
       .map((g) =>
-        Object.entries(variables.color['---light'][g]).map(([key, value]) => [
-          `${g}.${key}`,
-          new TinyColor(value.value).toHex8String()
-        ])
+        Object.entries(universalVariables.color['---light'][g]).map(
```
fallaciousreasoning, line 39: I really need to fix those prefixes in the variables exporter :laughing:

src/tokens/transformation/common/leoParser.js
```diff
@@ -30,12 +34,25 @@ module.exports = {
...
+            `${g}.${key}`,
+            new TinyColor(value.value).toHex8String()
+          ]
+        )
+      )
+      .flat()
+
+    const layerEffectColors = allowedGroups
+      .map((g) =>
+        Object.entries(layerVariables.color?.['---light']?.[g] || {}).map(
```
fallaciousreasoning, line 50: Maybe worth pulling this into a variable so when we change the name it only needs to be done in one place.

src/tokens/transformation/tailwind/formatFonts.js
```diff
@@ -16,8 +14,7 @@ module.exports = ({ dictionary }) => {
       fontClass += `-${t.attributes.state}`
     }
 
-    // Remove "desktop" and "browser" for typography
-    fontClass = stripTokenPrefix(fontClass)
+    fontClass = fontClass
```
fallaciousreasoning, line 17: nit: remove

src/tokens/transformation/common/leoParser.js
```diff
@@ -30,12 +34,25 @@ module.exports = {
     // export variable references in effects.
     // At the moment, we just whitelist a few color groups.
     const allowedGroups = ['elevation', 'primary', 'secondary']
-    const effectColors = allowedGroups
+    const universalEffectColors = allowedGroups
```
fallaciousreasoning, line 37: Maybe worth extracting this to a function we can reuse for `layerEffectColors` and `universalEffectColors`

src/tokens/transformation/common/leoParser.js
```diff
@@ -45,7 +62,9 @@ module.exports = {
         // Some entries are null in the style-dictionary export.
         if (!entry) continue
         const color = new TinyColor(entry.color).toHex8String()
-        const match = effectColors.find(([key, value]) => value === color)
+        const match =
```
fallaciousreasoning, line 65: Do you mind adding a comment about why we need to do this because I'm not going to remember :laughing:

fallaciousreasoning (approved):

src/tokens/transformation/common/leoParser.js
AlanBreck: I think abstracting to a function, as later suggested, solves this problem.

AlanBreck:
@fallaciousreasoning, output looks a little better, but seems that typography tokens in `variables.css` is still getting moved around?

fallaciousreasoning:
Interesting, I wonder why - looks like the new sort is slightly more accurate though

fallaciousreasoning (approved):
LGTM @AlanBreck - I was thinking about it some more, do you mind squashing this when you land? 

I'm thinking of it as a discrete tokens/variables update, so we can jump back to before/after this change. The parser changes are for this update, and don't necessarily make sense without it.

PR #555, opened 2024-02-15
Updated README.md

xylus11:
improved Readme for more readability and precison


fallaciousreasoning:
I think this is referring to the `tokens` part of the package, rather than the tokens themselves, so I don't think these changes are necessary.

PR #559, opened 2024-02-20
[Focus]: Set delegatesFocus on all components

fallaciousreasoning:
This delegates the focus to the first focusable element (or the element with `autofocus`) inside the shadowDOM when `.focus()` is called on the element.

I don't know if we want this for every component, but it does seem like a sensible default.

https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/delegatesFocus

### DIFF
--- a/src/components/svelte-web.ts
+++ b/src/components/svelte-web.ts
@@ -117,7 +117,8 @@ export default function registerWebComponent(
       // especially when being rendered inside a Polymer dom-if. In this case,
       // we need to also clear the contents of the node, to ensure we don't
       // duplicate content.
-      const shadow = this.shadowRoot ?? this.attachShadow({ mode })
+      const shadow =
+        this.shadowRoot ?? this.attachShadow({ mode, delegatesFocus: true })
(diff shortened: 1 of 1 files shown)


AlanBreck:
> which may be undesirable if that element is not meant to be part of the tabbing order (for example, an element with tabindex="-1"), or if a more 'important' focusable element should receive initial focus (for instance, the first text field rather than the 'close' button which precedes it).

This seems like a caveat worth consideration. What are the negatives of _not_ doing this?

fallaciousreasoning:
The negatives of not doing this are that for elements like the Button, Input, Dropdown ect calling `.focus()` on the HTMLElement doesn't behave as expected.

However, it sounds like we might not want this for non `input type="..."` components? I dunno.

fallaciousreasoning:
FWIW, @tapanmodh's PR https://github.com/brave/brave-core/pull/22220 depends on this.

Maybe we need to come up with a way for components to opt in/out of this behavior?

AlanBreck (approved):

PR #566, opened 2024-02-21
[Focus]: Make elements focusable, with a more limited scope than #559

fallaciousreasoning:
Closes #559 

With this approach, elements can opt in to being focusable, as opposed to in #559 where we opt everything in.

### DIFF
--- a/src/components/radioButton/radioButton.svelte
+++ b/src/components/radioButton/radioButton.svelte
@@ -48,6 +48,11 @@
     currentValue = value
     dispatch('change', { value })
   }
+
+  let input: HTMLElement | undefined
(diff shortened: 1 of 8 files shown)


src/components/radioButton/radioButton.svelte
```diff
@@ -48,6 +48,11 @@
     currentValue = value
     dispatch('change', { value })
   }
+
+  let input: HTMLElement | undefined
+  export function focus(options: FocusOptions) {
+    input?.focus()
```
fallaciousreasoning, line 54:
```suggestion
  export function focus(options?: FocusOptions) {
    input?.focus(options)
```

PR #578, opened 2024-03-08

brave-support-admin:

- has a raster image element
- has a raster image element
- has a raster image element


AlanBreck (approved):

AlanBreck (approved):

fallaciousreasoning:
@aguscruiz looks like we've added a bunch of different spinner states - we don't actually need these in the code, as we just spin the loading indicator. Do you think we could remove them?

aguscruiz:
> @aguscruiz looks like we've added a bunch of different spinner states - we don't actually need these in the code, as we just spin the loading indicator. Do you think we could remove them?

Yeah sure! Just moved all the ones that are not the main component to another page

AlanBreck (approved):

PR #588, opened 2024-03-13
[Toggle]: Fix sizing

fallaciousreasoning:


src/components/toggle/toggle.svelte
```diff
@@ -120,6 +121,7 @@
     &.size-small {
       --width: 40px;
       --height: 24px;
+      --leo-icon-size: 12px;
```
AlanBreck, line 124: Should this not be `--icon-size`?

fallaciousreasoning: Fixed - nice catch

petemill (approved):

PR #601, opened 2024-03-22

brave-support-admin:

- has a raster image element
- has a raster image element
- has a raster image element


fallaciousreasoning (approved):

icons/antena.svg
AlanBreck, line 1: Note: this will result in a breaking change.

@petemill / @fallaciousreasoning, are either of you aware of this being used in production?

fallaciousreasoning: Not that I know of - that said, there's a replacement, and we'll get a build failure in Brave-Core if something is using it.

PR #606, opened 2024-03-26
Add RTL support in Toggle and NavDots

fallaciousreasoning:
Currently, the Navdots and Toggle do some calculations with CSS properties which don't play nice with RTL. This fixes them.

### DIFF
--- a/src/components/navdots/navdots.svelte
+++ b/src/components/navdots/navdots.svelte
@@ -44,6 +44,14 @@
 </nav>
 
 <style lang="scss">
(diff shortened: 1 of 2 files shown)


AlanBreck (approved):
LGTM!

My only concern (without any ideas for a great solution) is that the values of `1` and `-1` don't correspond very nicely with `--leo-direction`.

fallaciousreasoning:
Yeah, not really sure what a better name would be :laughing:

PR #611, opened 2024-03-27
[FormItem]: Make it possible to customise control colors

fallaciousreasoning:
@aguscruiz could you take a look at this? For the Brave News stuff, I need to be able to customize the color of the `Input` control. To do this, I have to define the hover/background/border colors in terms of `--leo-control-color` mixed with the foreground/background color.

The colors are **slightly** different to what was there originally, so mind taking a look (and setting `--leo-control-color`). Looks nicest with colors which have some transparency (try `rgba(255,0,0,0.4)`).

### DIFF
--- a/src/components/formItem/formItem.svelte
+++ b/src/components/formItem/formItem.svelte
@@ -39,6 +39,11 @@
       description: 'The icon color used by the control',
       control: 'color'
     },
+    '--leo-control-color': {
+      description:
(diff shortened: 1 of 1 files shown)


aguscruiz:
@fallaciousreasoning I think they look great! Played around changing colors and couldn't break it, so that's a good sign 😂 Amazing!

PR #623, opened 2024-04-15
button menu's internal state cannot be overriden. work with a copy.

nullhook:


src/components/buttonMenu/buttonMenu.svelte
```diff
@@ -40,7 +41,7 @@
     <slot name="anchor-content">Click</slot>
   </div>
   <Menu
-    bind:isOpen
+    bind:isOpenInternal
```
fallaciousreasoning, line 44: I don't think you can bind `isOpenInternal` like this - what's it binding to?

fallaciousreasoning: Additionally, maybe we shouldn't use bind here at all? What if the API on Menu changes and it doesn't emit an onClose - `isOpenInternal` would change but our `isOpen` state wouldn't get updated (and we might not emit our close event).

nullhook: fixed

fallaciousreasoning (approved):
Looks good @nullhook (assuming you have it working in brave-core) - mind squashing when you commit and adding a `[ButtonMenu]: ` prefix, so we know what component this affected in the changelog.

PR #624, opened 2024-04-16
menu has user controlled height now

nullhook:


AlanBreck:
@nullhook, we've been somewhat careful to limit how many options we expose on each component. Can you explain more about why this is necessary?

nullhook:
image

i'm implementing a feature in aichat called 'slash tools' that anchors a button menu to the input box, allowing users to select an action. since the list can be lengthy, we plan to limit the maximum height of the button menu to prevent it from covering the chat conversation.

fallaciousreasoning (approved):
No objections on my end.

AlanBreck (approved):
LGTM as well.

@nullhook, when you `Squash and merge`, would you please format the commit message to the following?

```
[Component]: description of change starting with imperative verb
```

For instance it could be:

```
[ButtonMenu]: adds public css variable for setting max height
```

...or something along those lines.

PR #637, opened 2024-04-30
[Events]: Allow event handler objects

fallaciousreasoning:
I didn't realise this was a thing you could do, but if you pass an object with a `handleEvent` function, that will be invoked!

This is what `LitElement` does, so we won't work with LitElement until we fix this.

### DIFF
--- a/src/components/svelte-web.ts
+++ b/src/components/svelte-web.ts
@@ -254,16 +254,20 @@ export default function registerWebComponent(
 
     addEventListener(
       event: string,
-      callback: Callback,
+      eventHandler: EventListenerOrEventListenerObject,
       options?: boolean | AddEventListenerOptions
     ) {
       const svelteEvent = events[event]
       if (svelteEvent) {
+        const callback =
+          'handleEvent' in eventHandler
+            ? eventHandler.handleEvent.bind(eventHandler)
+            : eventHandler
         this[svelteEvent] = callback
         return
       }
 
-      super.addEventListener(event, callback, options)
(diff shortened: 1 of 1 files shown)


src/components/svelte-web.ts
```diff
@@ -254,16 +254,20 @@ export default function registerWebComponent(
...
-      callback: Callback,
+      eventHandler: EventListenerOrEventListenerObject,
       options?: boolean | AddEventListenerOptions
     ) {
       const svelteEvent = events[event]
       if (svelteEvent) {
+        const callback =
+          'handleEvent' in eventHandler
+            ? eventHandler.handleEvent.bind(eventHandler)
+            : eventHandler
```
AlanBreck, line 265: Should we being checking types here instead? Something like:

```js
if (typeof eventHandler === 'function') {
  callback = eventHandler
} else if ('handleEvent' in eventHandler) {
  callback = eventHandler.handleEvent.bind(eventHandler)
} else {
  throw Error("Must pass a function or include one in object as 'handleEvent'")
}
```

AlanBreck (approved):

src/components/svelte-web.ts
fallaciousreasoning: tbh, I'm not too fussed either way - this is pretty niche. Happy to change it, but I kinda like it as is.

AlanBreck: It's definitely more succinct as is. Let's go for it!

PR #651, opened 2024-05-03
[Tooltip]: Don't show tooltip unless there is content

fallaciousreasoning:
The content of the `Tooltip` isn't immediately available, so don't show the tooltip until we have content.

### DIFF
--- a/src/components/tooltip/tooltip.svelte
+++ b/src/components/tooltip/tooltip.svelte
@@ -31,7 +31,8 @@
   export let fadeDuration: number = 0
 
   /** Called when the visibility of the tooltip is changed */
-  export let onVisibilityChange: (detail: { visible: boolean }) => void = undefined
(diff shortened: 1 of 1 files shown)


src/components/tooltip/tooltip.svelte
```diff
@@ -31,7 +31,8 @@
   export let fadeDuration: number = 0
 
   /** Called when the visibility of the tooltip is changed */
-  export let onVisibilityChange: (detail: { visible: boolean }) => void = undefined
+  export let onVisibilityChange: (detail: { visible: boolean }) => void =
```
AlanBreck, line 34: Sometimes `prettier` is obnoxious.

AlanBreck (approved):

src/components/tooltip/tooltip.svelte
fallaciousreasoning: You can't tell prettier it aint prettier :D

PR #654, opened 2024-05-06
[ButtonMenu]: Make it possible to set the positionStrategy for ButtonMenu

fallaciousreasoning:
@AlanBreck we should find a way to specify that props should be forwarded :laughing:


AlanBreck (approved):
> @AlanBreck we should find a way to specify that props should be forwarded 😆

Seriously. Assuming for a moment that we didn't care about Svelte, is there a way of doing that in WC land that doesn't involve mutation observers?

fallaciousreasoning:
Sadly no, if we don't explicitly list out the props we support somewhere :frowning_face:

PR #659, opened 2024-05-07

brave-support-admin:

- has a raster image element
- has a raster image element
- has a raster image element


icons/browser-home.svg
AlanBreck, line 1: @aguscruiz, just confirming that the `*-home` icons are supposed to be redundant?

aguscruiz: Yeah, we still need both, but they're visually the same icon now

AlanBreck (approved):

PR #665, opened 2024-05-08
Button fixes

AlanBreck:
- **[Button]: removes background from plain-faint fab**
- **[Button]: prevents button content from shrinking**

### DIFF
--- a/src/components/button/button.svelte
+++ b/src/components/button/button.svelte
@@ -358,7 +358,7 @@
     }
   }
   .leoButton.isPlainFaint {
-    --foreground: transparent;
+    --foreground: var(--leo-color-black);
     --primary-color: currentColor;
     --color: var(--mixed-primary-color);
     --box-shadow-hover: none;
@@ -366,6 +366,7 @@
     --bg-hover-mix: 5%;
 
     @theme (dark) {
+      --foreground: var(--leo-color-white);
       --bg-hover-mix: 10%;
     }
 
@@ -374,6 +375,12 @@
         in srgb,
         var(--primary-color),
         var(--background) 30%
+      );
+
+      --icon-hover-color: color-mix(
+        in srgb,
+        var(--icon-color),
+        var(--foreground) 20%
       );
 
       position: relative;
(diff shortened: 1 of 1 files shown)


src/components/button/button.svelte
```diff
@@ -148,6 +148,7 @@
     color: var(--color);
     text-decoration: none;
     padding: var(--leo-button-padding, var(--padding-y) var(--padding-x));
+    min-width: max-content;
```
petemill, line 151: This doesn't mean that buttons won't ever reduce to fit inside their parent does it? Especially if in a flex or block-like full width state?

AlanBreck: It would. Do we really want the button contents stacking though?

fallaciousreasoning: I think sometimes we might (if I have two words it might be good to have it reflow to two lines on super small screens).

Out of interest, why do we need this? If it's just for the `fab` button icons, maybe we can add this to the `.fab` class?

src/components/button/button.svelte
```diff
@@ -156,6 +157,7 @@
 
     .content {
       padding: 0 var(--icon-gap);
+      flex-shrink: 0;
```
fallaciousreasoning, line 160: Should this be on `.fab` too? I feel like we want our buttons to be shrinkable

AlanBreck: > I feel like we want our buttons to be shrinkable

Do you mean we do or don't the `fab` buttons to shrink? `flex-shrink: 0` will prevent them from shrinking.

I wouldn't think we do since they have defined sizes.

src/components/button/button.svelte
AlanBreck: Went ahead and removed. It was a secondary matter for the PR, so we can pick up in future if needed.

src/components/button/button.svelte
fallaciousreasoning: Sorry, I meant the opposite - should that only be on the `fab` rather than on the normal buttons too.

like, should it be part of the `.fab` class :laughing: sorry for the ambiguity

fallaciousreasoning (approved):

PR #666, opened 2024-05-09
Upgrade Storybook and add themes addon

AlanBreck:
I've been encountering issues with the background color on stories randomly changing for no apparent reason. This PR aims to resolve this by leveraging the `@storybook/addon-themes` addon.

- **[Storybook]: upgrades to 8.x**
- **[Storybook]: changes theme switching to use @storybook/addon-themes**


.github/workflows/build.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       - name: Install dependencies
         run: |
           npm ci --ignore-scripts
+          npm explore esbuild -- npm run postinstall
```
AlanBreck, line 28: @thypon and/or @bcaller, is this acceptable? It's a workaround due to the following build error:

https://github.com/brave/leo/actions/runs/9215629603/job/25354366802#step:12:19

src/components/alert/alert.stories.svelte
```diff
@@ -91,8 +91,10 @@
   component={Alert}
   argTypes={{
     modes: { table: { disable: true } },
+    // @ts-ignore
```
AlanBreck, line 94: @fallaciousreasoning, I couldn't do `options: modes as any` since it was flagging `as` as an unknown token. I believe this is because TS isn't currently supported within the HTML portion of Svelte components? That's a bit of a guess, though.

Anyhow, figuring `@ts-ignore` works just as well? :-D

fallaciousreasoning: I think that's right, about `Typescript` not being supported in the HTML bit of `Svelte` components.

Hey, I just had another idea - can you pass through a copy of the props?

`modes={[...modes]}` or something?

.github/workflows/build.yml
thypon: looks good to me

src/components/alert/alert.stories.svelte
AlanBreck: That would probably have worked!

However, it appears that it's no longer an issue? Maybe because of a still more recent Storybook update I applied?

AlanBreck:
@fallaciousreasoning, mind reviewing when you get a chance?

.storybook/preview.ts
```diff
@@ -25,5 +25,15 @@ export const decorators = [
...
   },
-  () => Layout
+  () => Layout,
+  withThemeByDataAttribute({
+    themes: {
+      system: 'system',
+      light: 'light',
+      dark: 'dark'
+    },
+    defaultTheme: 'light',
```
fallaciousreasoning, line 35:
```suggestion
    defaultTheme: 'system',
```

I think this should default to the system theme

src/components/toggle/toggle.stories.svelte
```diff
@@ -49,7 +49,7 @@
       description: 'The color of the thumb when the control is disabled'
     },
     '--leo-toggle-label-gap': {
-      control: 'The space between the label and the toggle (in CSS units)'
```
fallaciousreasoning, line 52: lol

fallaciousreasoning (approved):

fallaciousreasoning (approved):

PR #671, opened 2024-05-14
[FormItem]: Make text color configurable

fallaciousreasoning:
We need this for the NTP search box


AlanBreck (approved):
LGTM!

Not blocking, but confirming two things:
1. It's actually intended that we should be allowed to customize this?
2. Should there be any updates to stories to reflect this additional CSS variable?

fallaciousreasoning:
Unfortunately, I think it needs to be - the NTP Search is forcing a different text color. The story automatically updates when you add something to the list of supported variables :smile:

PR #673, opened 2024-05-14
[Button]: deprecate fab prop in favor of iconOnly

AlanBreck:


fallaciousreasoning (approved):
Approving - this is a definite improvement, but do you mind holding off merging it for a couple of days - want to try and get the NTP Search stuff landed, and it might require Nala bumps (and this will make bumping Nala complicated :laughing: )

AlanBreck:
> but do you mind holding off merging it for a couple of days

Not at all.

fallaciousreasoning:
Just wondering out loud - could we change the prop to be `icon` so I can write

```tsx
<Button icon kind="outline">...</Button>
```

more wondering, but could it Just Work (tm) if you only slot in an Icon? Or is that too magic

AlanBreck:
So we decided to split this out into an `IconButton` instead of a prop, yes? cc @fallaciousreasoning @petemill

PR #680, opened 2024-05-18

brave-support-admin:

- has a raster image element
- has a raster image element
- has a raster image element


AlanBreck:
@aguscruiz, the changes to the icon colors seem pretty significant. Should these proceed to production, or do we need to wait to land this until we have corresponding color token updates?

aguscruiz:
@AlanBreck safe to ignore, it's all because of the new tokens, they changed the icon's default color so that's why it's prompting it

fallaciousreasoning:
@AlanBreck maybe we should always set the color on the icons to whatever value we have for `--leo-color-icon-default` or something, so that color changes in Figma don't trigger icon changes?

AlanBreck:
> @AlanBreck maybe we should always set the color on the icons to whatever value we have for `--leo-color-icon-default` or something, so that color changes in Figma don't trigger icon changes?

@fallaciousreasoning, are you saying we could do that within our action/script for pulling icon changes?

PR #686, opened 2024-05-23
[React]: Add slot support

fallaciousreasoning:
This lets you set the `slot` attribute directly on Nala components

### DIFF
--- a/src/components/svelte-react.ts
+++ b/src/components/svelte-react.ts
@@ -38,6 +38,7 @@ type Ref<T> =
     }
 export type ReactProps<Props> = Props & {
   ref?: Ref<Props>
+  slot?: string
 } & {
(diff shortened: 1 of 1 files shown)


src/components/svelte-react.ts
```diff
@@ -38,6 +38,7 @@ type Ref<T> =
     }
 export type ReactProps<Props> = Props & {
   ref?: Ref<Props>
+  slot?: string
```
AlanBreck, line 41: Does `ref?: Ref<Props>` not need a comma?

fallaciousreasoning: It was `npm run format` which removed it

But also, nope :laughing: In Ts, valid separators are `\n`, `,` and `;` 

You can even mix/match as you see fit, if you're a monster.

AlanBreck (approved):

PR #689, opened 2024-05-23
Ensure consuming apps receive Icon's IconName type instead of a bad relative import path

petemill:
Consuming apps should know @brave/leo but ../../../ is not the correct relative path from the typescript build output path

I confirmed I didn't need to change anything for consuming apps to get the intellisense to work:
image

### DIFF
--- a/examples/example-ui-react/src/App.tsx
+++ b/examples/example-ui-react/src/App.tsx
@@ -1,13 +1,13 @@
 import * as React from 'react'
-import '../../../tokens/css/variables.css'
-import LeoButton from '../../../react/button'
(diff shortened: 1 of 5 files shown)


fallaciousreasoning:
Just want to double check - that screenshot is in Brave Core?

fallaciousreasoning (approved):

petemill:
> Just want to double check - that screenshot is in Brave Core? Also, when you land it, could you prefix the commit with `[Icon]:` so when I do the release notes its easy 😄

Yes, it's from brave-core

Issue #690, opened 2024-05-26
Button Web Component doesn't treat "isDisabled" as a boolean type


petemill:
This does not result in a disabled button:
```
<leo-button id="clickable" isDisabled>Disabled button</leo-button>
```

This does:
```
<leo-button id="clickable" isDisabled="anything">Disabled button</leo-button>
```

When telling a rendering framework like Polymer that the attribute is a boolean attribute, it will add the attribute (without a string value). This does not result in a disabled button.
Likewise when telling the rendering framework it's simply a string attribute, all buttons will be disabled, since "true" or "false" result in `isDisabled` being present.

fallaciousreasoning:
:facepalm: I bet we're checking `!isDisabled` or something, and because the empty string is falsey....

PR #691, opened 2024-05-26
[Button]: Make empty string truthy in HTML attributes

fallaciousreasoning:
Fixes https://github.com/brave/leo/issues/690

### DIFF
--- a/src/components/button/button.svelte
+++ b/src/components/button/button.svelte
@@ -51,6 +51,7 @@
   export let onClick: () => void = undefined
 
   $: tag = href ? 'a' : ('button' as 'a' | 'button')
+  $: disabled = !!(isDisabled || (isDisabled as any) === '')
 </script>
 
 <svelte:element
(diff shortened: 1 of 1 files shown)


src/components/button/button.svelte
```diff
@@ -43,14 +45,19 @@
...
   export let fab = false
 
   export let onClick: () => void = undefined
 
   $: tag = href ? 'a' : ('button' as 'a' | 'button')
+
+  // Web components set these attributes to the empty string, which is falsey
+  // so we need to consider the empty string as truthy.
+  $: loading = !!(isLoading || isLoading === '')
+  $: disabled = !!(isDisabled || isDisabled === '')
```
petemill, line 60: Is this unfortunate that we need to make this consideration for every boolean prop on every component? Can we abstract it away?

petemill (approved):
nit: can we do the non-relevant formatting at least in a separate commit?

src/components/button/button.svelte
fallaciousreasoning: There probably is - unfortunately it'd require us pulling out the type of the prop at build time and treating them differently in the webcomponents wrapper. 

Doable, but probably not super quick.

I had a bit more of a think about why we haven't encountered this before:

At runtime, we check the value of the props and if they're a boolean, we treat that attribute as a boolean from then on. Unfortunately, `isDisabled` is `undefined` by default.

I've removed the wrapping for isLoading, based on this.

PR #694, opened 2024-05-27
[Icon]: Add support for libraries of icons

fallaciousreasoning:
The motivation for this is I need to be able to display Chromium icons from `chrome://resources/images/*` on the NTP alongside Nala icons.

Previously, I was just displaying the icon via an ``. However, when I do this, I can't change to color. Fortunately, Nala supports letting me do this sort of thing already, so if I can add an Icon library for the upstream icons, this should work nicely :smile:

### DIFF
--- a/src/components/icon/icon.svelte
+++ b/src/components/icon/icon.svelte
@@ -3,12 +3,22 @@
   import { writable } from 'svelte/store'
 
   let lastIconBasePath = '/icons'
-  let iconBasePath = writable(lastIconBasePath)
+  let iconBasePath = writable<{
+    [key: string]: string
+  }>({
+    '': lastIconBasePath
+  })
 
   export const setIconBasePath = (basePath: string) => {
     lastIconBasePath = basePath
-    iconBasePath.set(basePath)
+    addLibrary('', basePath)
   }
+
+  export const addLibrary = (library: string, basePath: string) =>
+    iconBasePath.update((prev) => ({
+      ...prev,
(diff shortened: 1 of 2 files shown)


src/components/icon/icon.svelte
```diff
@@ -3,13 +3,23 @@
   import { writable } from 'svelte/store'
 
   let lastIconBasePath = '/icons'
-  let iconBasePath = writable(lastIconBasePath)
+  let iconBasePath = writable<{
```
AlanBreck, line 6: `iconBasePath` feels like the wrong name for this data structure. Maybe the following changes?
- `lastIconBasePath => defaultLibraryPath`
- `iconBasePath => iconLibraryPaths`

src/components/icon/icon.svelte
```diff
@@ -3,13 +3,23 @@
...
+    [key: string]: string
+  }>({
+    '': lastIconBasePath
```
AlanBreck, line 9: Any way we could use something like `default` instead of an empty string?
```suggestion
    DEFAULT: lastIconBasePath
```

fallaciousreasoning: We could do something like:

```
const defaultSymbol = new Symbol()

{
    [defaultSymbol]: lastIconBasePath
}
```

fallaciousreasoning:
Closing this as the idea still scares me a wee bit :laughing:

Issue #682, opened 2024-05-20
List of component changes with Material color update


aguscruiz:
```[tasklist]
### Tasks
- [ ] Label colors. Default backgrounds change from themable 10 to 20.
- [ ] Toggle colors change, all of them.
- [ ] All themable colors and semantic colors change to adapt to new primitives
- [ ] Buttons change and now have light-dark themes in filled buttons
- [ ] Default icon color changes
- [ ] Default text styles color changes
- [ ] IT would be ideal if we could add aliases for each color that references another. Would make it easier to change things in the future?
- [ ] Updated elevation values
- [ ] Updated focus state: Created one style for buttons and one style for everything else as a stroke variable
```

PR #695, opened 2024-05-29
[Tokens]: Update universal with Material Base

fallaciousreasoning:
Resolves https://github.com/brave/leo/issues/682

## TODO
- [x] Change button text color in dark mode
- [x] Fix toggle dark mode
- [x] Fix radiobutton dark mode
- [x] Fix checkbox dark mode


aguscruiz:
What if instead of Static I use something like "Default"? 
-----
Yes, Android and iOS go in Universal too now.
-----
Is there any way we can see where does primitive colors were used? I'm not concerned about themable, but primitives will break
-----
There's something else I added over the last days, and it's the "blurple" and "orange" colors, in case we ever need to use them as they are, even if the theme is dynamic.

fallaciousreasoning:
@aguscruiz wrt to static/default I actually got it working
same deal with the primitive colors - I think things should just work - gotta say this actually works really well!

fallaciousreasoning:
Okay, I think this might be ready for a first review :fearful: @petemill @AlanBreck 

## Things I'd like to do:
1. Get reference resolving working with Style Dictionary - this seems to make the export work pretty darn well but the format that the variables needs to be in is a bit odd (we need to introduce a top level object to make it work).

aguscruiz:
@fallaciousreasoning just wanted to double check something (out of ignorance)

With this update, would we get the advantage of aliasing tokens we talked about?

The one where for example, `text/primary` would ve defined as `themable/neutral/70` (or whatever it s) instead of the actual pure hex code?

fallaciousreasoning:
Yup, that's right @aguscruiz - I changed the way we're doing it slightly (with a new `aliases` field) so we don't break the Android/iOS tokens. 

The stuff in the `value` field is essentially the default `Static` value while `aliases` tells it what variable it should point to

aguscruiz:
Yayyy thanks ⭐⭐⭐⭐⭐

src/components/button/button.svelte
```diff
@@ -389,15 +386,15 @@
     }
   }
   .leoButton.isPlainFaint {
-    --foreground: var(--leo-color-black);
+    --foreground: var(black);
```
fallaciousreasoning, line 389:
```suggestion
    --foreground: black;
```

src/tokens/transformation/web/color.js
```diff
@@ -0,0 +1,20 @@
+const TinyColor = require('@ctrl/tinycolor')
+
+module.exports = {
+  type: 'value',
+  matcher(token) {
+    return token.type === 'color'
+  },
+  transformer({ value, referencedVariable }) {
+    if (referencedVariable) {
+      const name = referencedVariable.substring(1).replaceAll('.', '-')
```
petemill, line 10: Small point: This is the same as what we do with skia for referenced variables too, right (ignore the first char, and replace `.` with `-`? Should the knowledge of how the references variables translate to names be shared?

petemill (approved):
Amazing to get this done - thanks for taking it on

fallaciousreasoning:
Merging... I'm scared :fearful:

PR #698, opened 2024-06-01
[Storybook]: ensures layer tokens are only loaded when viewing story

AlanBreck:
Prior to this PR, tokens that get overriden by other layers were not accurately being displayed in Storyook. See for example font tokens for `Browser`.

### DIFF
--- a/src/tokens/browserFontTokens.stories.svelte
+++ b/src/tokens/browserFontTokens.stories.svelte
@@ -1,8 +1,18 @@
 <script lang="ts">
+  import { onMount } from 'svelte'
   import { Meta, Story } from '@storybook/addon-svelte-csf'
   import { font as allFonts } from '../../tokens/css/variables-browser'
   import FontTokenSwatchGroup from '../storyHelpers/FontTokenSwatchGroup.svelte'
-  import '../../tokens/css/variables-browser.css'
+  // @ts-ignore
+  import styles from '../../tokens/css/variables-browser.css?raw'
+
+  onMount(() => {
+    const stylesheet = new CSSStyleSheet();
+    stylesheet.replaceSync(styles);
+    document.adoptedStyleSheets = [stylesheet];
+
+    return () => document.adoptedStyleSheets = [];
+  })
 </script>
 
 <Meta title="Tokens/Browser/Fonts" />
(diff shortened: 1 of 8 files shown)


src/tokens/browserFontTokens.stories.svelte
```diff
@@ -1,8 +1,18 @@
...
+  // @ts-ignore
+  import styles from '../../tokens/css/variables-browser.css?raw'
+
+  onMount(() => {
+    const stylesheet = new CSSStyleSheet();
+    stylesheet.replaceSync(styles);
+    document.adoptedStyleSheets = [stylesheet];
+
+    return () => document.adoptedStyleSheets = [];
+  })
```
AlanBreck, line 15: Thanks for introducing me to `adoptedStylesheets`, @fallaciousreasoning!

src/tokens/newtabFontTokens.stories.svelte
```diff
@@ -1,8 +1,18 @@
 <script lang="ts">
+  import { onMount } from 'svelte'
   import { Meta, Story } from '@storybook/addon-svelte-csf'
   import { font as allFonts } from '../../tokens/css/variables-newtab'
   import FontTokenSwatchGroup from '../storyHelpers/FontTokenSwatchGroup.svelte'
-  import '../../tokens/css/variables-newtab.css'
+  // @ts-ignore
+  import styles from '../../tokens/css/variables-newtab.css?raw'
+
+  onMount(() => {
```
fallaciousreasoning, line 9: Bit annoying that we have to do this separately for everything :/

fallaciousreasoning (commented):
Sucks we have to do it for each story separately :cry: 

What if we added a function like:

```ts
async function scopedApplyLayer(name: string) {
    const css = await fetch(`/tokens/css/${name}.css`).then(r => r.text())
    const ss= new CSSStyleSheet()
    ss.replaceSync(css)
    document.adoptedStyleSheets.push(ss)

    return () => {
        document.adoptedStyleSheets = document.adoptedStyleSheets.filter(s => s !== ss)
    }
}

src/tokens/newtabFontTokens.stories.svelte
AlanBreck: Had to tweak your suggested function since `onMount` can't be async AND return a function for unmount, but here's what it looks like:

https://github.com/brave/leo/pull/698/commits/871182b96465413e6fa694a2c620a92114f72b49

Pretty minor, but one down side here is that the css files get fetched for each story. Benefit probably outweighs the negative here, though.

fallaciousreasoning (approved):

PR #704, opened 2024-06-05
[Icons]: Don't change color on monochrome icons in updates

fallaciousreasoning:
This PR has two commits:
1. Updates the script to not trigger updates when the monochrome color changes
2. Updates the icons

They should probably be rebased rather than squashed so the icon updates commit only updates icons (finding things in them is a real pain)


fallaciousreasoning:
@AlanBreck what do you think of this approach?

AlanBreck (commented):
LGTM! ~Not exactly following what you're referring to with rebasing, though.~ You mean that this PR should be rebased instead of squashed, yes? For whatever reason, I thought you were referring to some change going forward with how we handle icon updates.

AlanBreck (approved):

PR #736, opened 2024-07-03
Ensure consuming apps receive Icon's IconName type instead of a bad relative import path

petemill:
Consuming apps should know @brave/leo but ../../../ is not the correct relative path from the typescript build output path

I had this pending locally but can't remember if it's still needed, so I'm submitting whilst I take a look!


fallaciousreasoning (approved):
LGTM - I thought we already merged this ages ago :D

PR #742, opened 2024-07-15
[Dropdown]: Update spacing and padding of dropdown options

aguscruiz:
I updated the padding of the items inside the dropdown, also the border radius and spacing between elements.

### DIFF
--- a/src/components/dropdown/dropdown.stories.svelte
+++ b/src/components/dropdown/dropdown.stories.svelte
@@ -207,7 +207,7 @@
   .country {
     display: flex;
     flex-direction: row;
-    gap: 8px;
+    gap: var(--leo-spacing-xl);
     align-items: center;
   }
 
--- a/src/components/menu/menu.svelte
+++ b/src/components/menu/menu.svelte
@@ -261,7 +261,7 @@
     cursor: pointer;
     margin: var(--leo-menu-item-margin);
     border-radius: var(--leo-menu-item-border-radius);
-    padding: var(--leo-menu-item-padding)
+    padding: var(--leo-menu-item-padding);
     display: revert;
   }
 
--- a/src/components/menu/menu.svelte
+++ b/src/components/menu/menu.svelte
@@ -242,8 +242,7 @@
   :global .leo-menu-popup ::slotted(*),
   :global .leo-menu-popup > * {
     --leo-menu-item-margin: var(--leo-spacing-s);
-    --leo-menu-item-padding-horizontal: var(--leo-spacing-xl);
-    --leo-menu-item-padding-vertical: var(--leo-spacing-m);
+    --leo-menu-item-padding:  var(--leo-spacing-m) var(--leo-spacing-xl);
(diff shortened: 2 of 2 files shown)


src/components/dropdown/dropdown.stories.svelte
```diff
@@ -207,7 +207,7 @@
   .country {
     display: flex;
     flex-direction: row;
-    gap: 8px;
+    gap: var(--leo-spacing-xl);
```
fallaciousreasoning, line 210: **Note:** This only affects the story, not the component - probably good for demonstration purposes :smile:

src/components/menu/menu.svelte
```diff
@@ -243,8 +242,9 @@
   :global .leo-menu-popup ::slotted(*),
   :global .leo-menu-popup > * {
     --leo-menu-item-margin: var(--leo-spacing-s);
-    --leo-menu-item-padding: var(--leo-spacing-xl);
-    --leo-menu-item-border-radius: var(--leo-spacing-m);
+    --leo-menu-item-padding-horizontal: var(--leo-spacing-xl);
+    --leo-menu-item-padding-vertical: var(--leo-spacing-m);
```
fallaciousreasoning, line 246: Don't think we need to separate that into two separate variables:

```suggestion
    --leo-menu-item-padding: var(--leo-spacing-m) var(--leo-spacing-xl);
```

This variable is just so people using the component can tweak it :smile:

src/components/menu/menu.svelte
```diff
@@ -261,7 +261,7 @@
     cursor: pointer;
     margin: var(--leo-menu-item-margin);
     border-radius: var(--leo-menu-item-border-radius);
-    padding: var(--leo-menu-item-padding);
+    padding: var(--leo-menu-item-padding-vertical) var(--leo-menu-item-padding-horizontal);
```
fallaciousreasoning, line 264:
```suggestion
    padding: var(--leo-menu-item-padding)
```

fallaciousreasoning (commented):
Thanks @aguscruiz !

src/components/menu/menu.svelte
aguscruiz: @fallaciousreasoning I accepted your suggestions there but it broke for me

image

aguscruiz:
@fallaciousreasoning I wanted to change the gap here between text and icon to `--leo-spacing-xl` but I couldn't manage to do it :( 

image

fallaciousreasoning (approved):

fallaciousreasoning:
Just confirming:

Custom options don't seem to line up with the `select...` text:
image


No icons have more padding on the side - it this what we want?
image

aguscruiz:
Yeah, it's the way I have it in Nala. Problem is a spacing value I have for the dropdown elements.

But let's keep it that way for now with this changes, and then I'll see if I can solve it. Changing one of these values is gonna have a ripple effect in Figma XD 

image

AlanBreck (approved):

PR #743, opened 2024-07-16
[Input]: Make input first focused element

fallaciousreasoning:
Currently, if the left-icon is interactive, that will be focused before the input


AlanBreck (commented):
@fallaciousreasoning, out of curiosity, why would the left icon need to be interactive?

fallaciousreasoning:
On the NTP Search box it opens a menu:
image

AlanBreck (approved):

PR #744, opened 2024-07-18

brave-support-admin:

- has a raster image element
- has a raster image element
- has a raster image element


AlanBreck (approved):

icons/text-input.svg
AlanBreck, line 1: @aguscruiz, is it intended that this should be removed?

aguscruiz: Yeah, we have another one `autofill` that's similar but more up to date in terms of style. I don't think anyone was using the other one

AlanBreck: @fallaciousreasoning, are you able to confirm on the browser side?

PR #746, opened 2024-07-22
Added a quick animation for hover state in inputs

aguscruiz:
Added a transition for box shadow and border in inputs

### DIFF
--- a/src/components/formItem/formItem.svelte
+++ b/src/components/formItem/formItem.svelte
@@ -240,7 +240,7 @@
     background: var(--background);
     box-shadow: var(--shadow);
     border: 1px solid var(--border-color);
-    transition: box-shadow 0.15s ease-in-out, border-color 0.15s ease-in-out;
+    transition: box-shadow 0.2s ease-in-out, border-color 0.2s ease-in-out;
 
 
     cursor: pointer;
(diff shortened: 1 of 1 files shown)


src/components/formItem/formItem.svelte
```diff
@@ -240,6 +240,8 @@
     background: var(--background);
     box-shadow: var(--shadow);
     border: 1px solid var(--border-color);
+    transition: box-shadow 0.15s ease-in-out, border-color 0.15s ease-in-out;
```
fallaciousreasoning, line 243: I think in other places we use 0.12s or 0.2s depending on the animation. Probably not super important though

fallaciousreasoning (approved):

src/components/formItem/formItem.svelte
AlanBreck: Might be worth creating a token for transitions.

AlanBreck (commented):
Apart from @fallaciousreasoning's comment about consistency, this looks good.

aguscruiz:
@AlanBreck I do have some presets for animations that we can add as tokens
https://www.figma.com/design/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?m=auto&t=3IMNIFI3ukRfbOE6-1

src/components/formItem/formItem.svelte
aguscruiz: Nice, thanks, updated to 0.2

AlanBreck (approved):

PR #747, opened 2024-07-23
[Typescript]: Add typescript support to scripts/token parsing

fallaciousreasoning:
This adds support for progressively converting tokens and scripts to Typescript via use of `tsx` (ts execute, similar to `ts-node` or `npx` but works with modules).

Unlike `ts-node`, `tsx` supports `esmodules`, so you can write typescript as normal!
Unlike `npx`, `tsx` supports typescript!

I was inspired by someone in brave-core asking whether we could start converting our scripts there to `tsx`

This allows us to progressively convert `.js` files to `.ts` - this should be super handy when messing around with stuff in `leoParser.js`

### DIFF
--- a/src/scripts/update-icons.ts
+++ b/src/scripts/update-icons.ts
@@ -1,9 +1,9 @@
-const figmaApiExporter = require('figma-api-exporter').default
-const fs = require('fs')
-const path = require('path')
-const { JSDOM } = require('jsdom')
-const { optimize } = require('svgo')
-const { TinyColor } = require('@ctrl/tinycolor')
+import figmaApiExporter from 'figma-api-exporter'
+import fs from 'fs'
+import path from 'path'
+import { JSDOM } from 'jsdom'
+import { optimize } from 'svgo'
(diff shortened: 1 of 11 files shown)


src/scripts/update-icons.ts
```diff
@@ -88,12 +88,12 @@ const mutateIcon = (iconName) => {
   if (!isProbablyColor(iconName)) {
     const idealColor = '#62757E'
     for (const filled of document.querySelectorAll('[fill]')) {
-      if (isAllowedColor(filled.getAttribute('fill'))) continue
+      if (isAllowedColor(filled.getAttribute('fill')!)) continue
```
AlanBreck, line 91: Out of curiosity, what is the `!` operator doing here?

AlanBreck (approved):
This will be great!

src/scripts/update-icons.ts
fallaciousreasoning: It means "this is not null/undefined"

PR #749, opened 2024-07-23
[Tokens](Tailwind): Only wrap shadow colors in function if necessary

AlanBreck:
The fix in https://github.com/brave/leo/pull/699 was too greedy, so this
adds additional checks to ensure that the color function isn't double
wrapped.

### DIFF
--- a/src/tokens/transformation/tailwind/twShadows.ts
+++ b/src/tokens/transformation/tailwind/twShadows.ts
@@ -21,7 +21,9 @@ export default {
  */
 const formatColorForTw = (colorString: string) => {
   const formattedColor = formatColor('tw', colorString)
-  return colorString.startsWith('$')
+  return colorString.startsWith('$') &&
+    !formattedColor.includes('elevation') &&
+    !formattedColor.startsWith('rgb')
     ? `rgba(${formattedColor}, 1)`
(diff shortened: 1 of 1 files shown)


src/tokens/transformation/tailwind/twShadows.js
```diff
@@ -20,7 +20,9 @@ module.exports = {
  */
 const formatColorForTw = (colorString) => {
   const formattedColor = formatColor('tw', colorString)
-  return colorString.startsWith('$')
+  return colorString.startsWith('$') &&
+    !formattedColor.includes('elevation') &&
+    !formattedColor.startsWith('rgb')
```
fallaciousreasoning, line 25: We don't do this for hex colors?

fallaciousreasoning (approved):
lgtm % nit

src/tokens/transformation/tailwind/twShadows.js
AlanBreck: Hmm... good question. I don't believe we're using HEX in Tailwind, though.

fallaciousreasoning: Probably not a super useful check at any rate

PR #754, opened 2024-07-26
[SegmentedControl]: Update styles for pill background and hover background

aguscruiz:
There was an issue that surfaced in wallet, where the segmented control had very low contrast. I checked and it was different from figma, so here's an update

- themable/10 for component background
- container/highlight for hover state

image

### DIFF
--- a/src/components/segmentedControl/segmentedControl.svelte
+++ b/src/components/segmentedControl/segmentedControl.svelte
@@ -111,7 +111,7 @@
 
   .leo-segmented-control {
     --leo-icon-size: var(--leo-icon-m);
-    --bg: var(--leo-color-gray-10);
+    --bg: var(--leo-color-neutral-10);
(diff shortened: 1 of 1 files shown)


AlanBreck (approved):

AlanBreck (changes requested):
I approved too quickly. It's complaining that `--leo-color-gray-10` doesn't exist...

AlanBreck (approved):

fallaciousreasoning:
Is this intentional? There doesn't seem to be a background in light mode?
image

fallaciousreasoning (approved):

aguscruiz:
@fallaciousreasoning there is, but the preview panel background is the same. I was talking with @AlanBreck about it, seems that the pannel is using page/background for bg instead of container/background. He has a fix

PR #757, opened 2024-07-29
[Alert]: Fix the padding on the Alert when removing the icon

fallaciousreasoning:
This fixes an issue which @simonhong encountered where hiding the icon still shows the spacing for it.

### DIFF
--- a/src/components/alert/alert.svelte
+++ b/src/components/alert/alert.svelte
@@ -21,6 +21,7 @@
   export let isToast = false
   export let hasActions = $$slots.actions
   export let hasContentAfter = $$slots['content-after']
+  export let hideIcon = false
 
   $: currentType = type ?? 'error'
   $: currentMode = mode ?? 'simple'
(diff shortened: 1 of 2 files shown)


petemill (approved):

src/components/alert/alert.svelte
```diff
@@ -21,6 +21,7 @@
   export let isToast = false
   export let hasActions = $$slots.actions
   export let hasContentAfter = $$slots['content-after']
+  export let hideIcon = false
```
petemill, line 24: will it show on the story?

src/components/alert/alert.svelte
```diff
@@ -93,6 +96,8 @@
     & .icon {
       --leo-icon-size: var(--leo-icon-m);
       color: var(--leo-icon-color);
+
```
AlanBreck, line 99: uber nit: why the space here?

src/components/alert/alert.svelte
fallaciousreasoning: Yup, props are automatically in the story :smile:

PR #766, opened 2024-08-01
[Tokens] Updated styles file for universal

aguscruiz:
Latest style set from Figma

- Font styles
- Shadow colors

### DIFF
--- a/src/tokens/universal.tokens.json
+++ b/src/tokens/universal.tokens.json
@@ -62,11 +62,11 @@
           "stops": [
             {
               "position": 0,
-              "color": "#3b08c2ff"
+              "color": "#4641eeff"
             },
             {
               "position": 1,
-              "color": "#3f39e8ff"
+              "color": "#6261ffff"
             }
           ]
         },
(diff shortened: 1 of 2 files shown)


src/tokens/universal.tokens.json
```diff
@@ -1198,37 +970,6 @@
     }
   },
   "effect": {
-    "focus state": {
```
fallaciousreasoning, line 1201: Could we keep these - otherwise updating is going to be painful in brave-core :smile:

fallaciousreasoning:
@AlanBreck - I actually didn't need to update the exporter - I locally saved the wrong file :facepalm: 

The font tokens are already in the `universal.variables.json` file - I don't know if we do anything with them yet, but do you mind if we merge this as-is?

fallaciousreasoning (approved):

AlanBreck:
> @AlanBreck - I actually didn't need to update the exporter - I locally saved the wrong file 🤦
> 
> The font tokens are already in the `universal.variables.json` file - I don't know if we do anything with them yet, but do you mind if we merge this as-is?

That's fine with me!

PR #768, opened 2024-08-05
[Tokens]: Update variables file with correct names for typography tokens

fallaciousreasoning:


fallaciousreasoning:
Sorry @AlanBreck it's got a new ordering it likes for the variables file but the output is more correct now - hopefully this makes it easier to use the variants you need :smile:

AlanBreck (approved):

PR #770, opened 2024-08-07
[ButtonMenu]: Don't have default anchor slot

fallaciousreasoning:


src/components/buttonMenu/buttonMenu.stories.svelte
```diff
@@ -28,6 +28,7 @@
 <Template let:args>
   <div class="container">
     <ButtonMenu>
+      <span slot="anchor-content"> Click Me! </span>
```
AlanBreck, line 31: Why the spaces around ` Clicke Me! `?

AlanBreck (approved):

src/components/buttonMenu/buttonMenu.stories.svelte
fallaciousreasoning: I think the formatter did it - I originally had it split across a couple of lines - I removed them and its still happy though :smile:

PR #775, opened 2024-08-08
[Skia]: Update how colors work to make them easier to use with Chromium's ColorProvider

fallaciousreasoning:
This is with the goal of being able to do

```cpp
auto mixer = provider->AddMixer();
mixer[kToolbarBackground] = {nala::kColorNeutral100000000};
```

additionally, it will let us base the Nala tokens off the Material tokens already in Chromium and let us support Material themes in Nala C++


fallaciousreasoning:
@simonhong and @sangwoo108 could you guys give me some feedback on this (as you're the experts :laughing: )

sangwoo108 (commented):
I believe all of us were waiting for this! Can't wait 😃

src/tokens/transformation/skia/index.ts
```diff
@@ -46,6 +46,30 @@ const getSortKey = ({ name, value }) => {
...
+
+const dynamicPalettePrimitives = [
+  'primary',
+  'secondary',
+  'tertiary',
+  'neutral',
+  'neutralVariant',
+  'error'
+]
+const maybeDynamicProps = (token: TransformedToken) => {
```
sangwoo108, line 58: Can we have a comment that describes what would expected inputs be like?

sangwoo108 (approved):
LGTM 👍

src/tokens/transformation/skia/templates/nala_color_mixer.cc.template
```diff
@@ -0,0 +1,36 @@
...
+#include "third_party/skia/include/core/SkColor.h"
+#include "ui/color/color_mixer.h"
+#include "ui/color/color_provider.h"
+#include "ui/color/color_provider_key.h"
+#include "ui/color/color_recipe.h"
+#include "ui/color/color_transform.h"
+
+namespace nala {
+
+void AddNalaColorMixer(ui::ColorProvider* provider,
```
simonhong, line 18: IMO, it would be great if nala doesn't have dependency on brave-core as much as possible.
It also could be helpful during the upstream bump. WDYT?

fallaciousreasoning: So yeah, I went back and forward on this a bit - at the end of the day though, we're trying to generate outputs from Nala that make them as easy as possible to use on different platforms (in my head this is the `Chromium Views` target).

The thing which tipped me over to doing it this way was that this file needs to be generated, and this is where we know all the relationships between the different variables. We could do this as part of the brave-core build process.

It's a good point about upstream bumps though, I hadn't thought of that.

Be interested to hear what @petemill and @AlanBreck think though

simonhong: Right, it should be generated from here. I also couldn't find better way.
Anyway, this PR is great 👍🏼

simonhong (approved):
++ 👍🏼

src/tokens/transformation/skia/index.ts
```diff
@@ -46,6 +46,30 @@ const getSortKey = ({ name, value }) => {
...
+  if (!token.name.includes('primitive')) return
```
AlanBreck, line 59: Sorry if I'm totally confused here, but shouldn't this return if the name _does_ include `primitive` as primitives are not dynamic?

AlanBreck (commented):
One question, but otherwise LGTM.

src/tokens/transformation/skia/index.ts
fallaciousreasoning: I've renamed this function to `maybeMaterialColorProps` which is a better name - what this function is doing is mapping the Material color primitives to their equivalent from the Chromium color mixer.

AlanBreck (approved):

PR #776, opened 2024-08-12
[ButtonMenu]: **BREAKING** Fix `min-width` variable and close via click outside

fallaciousreasoning:


src/components/buttonMenu/buttonMenu.svelte
```diff
@@ -31,6 +31,11 @@
     if (isOpen === undefined) isOpenInternal = toggleTo
     onChange?.({ isOpen: toggleTo })
   }
+
+  const handleClose = (e: CloseEventDetail) => {
+    if (isOpen === undefined) isOpenInternal = false
+    onClose?.(e)
+  }
```
nullhook, line 38: the state wasn't being updated onClose and therefore the menu took multiple clicks to open again?

src/components/menu/menu.svelte
```diff
@@ -223,6 +223,8 @@
   }
 
   .leo-menu .leo-menu-popup {
+    --min-width: var(--leo-menu-control-min-width, var(--leo-menu-target-width));
```
nullhook, line 226: brave rewards codebase uses the previous name so we will need to change it to this when updating nala

src/components/buttonMenu/buttonMenu.svelte
fallaciousreasoning: Exactly, yeah :smile:

src/components/menu/menu.svelte
fallaciousreasoning: Good to know!

src/components/buttonMenu/buttonMenu.svelte
```diff
@@ -31,6 +31,11 @@
     if (isOpen === undefined) isOpenInternal = toggleTo
     onChange?.({ isOpen: toggleTo })
   }
+
+  const handleClose = (e: CloseEventDetail) => {
+    if (isOpen === undefined) isOpenInternal = false
+    onClose?.(e)
```
josheleonard, line 37: Should this also fire the onChange event?

fallaciousreasoning: Nice catch! That's probably why the controlled version wasn't working!

src/components/buttonMenu/buttonMenu.svelte
```diff
@@ -31,6 +31,12 @@
...
+    onChange?.({ isOpen })
```
josheleonard, line 38: How about forcing the event value to `false` here?
    onChange?.({ isOpen: false })
```

**Comment by AlanBreck** [comment]
Since this change would be breaking, what's the migration path for existing consumers?

File: src/components/buttonMenu/buttonMenu.stories.svelte [hunk 5]
```diff
@@ -16,11 +16,13 @@
   title="Components/ButtonMenu"
   component={ButtonMenu}
   argTypes={{
-    '--leo-menu-control-width': {
-      description: '(readonly): Computed width of menu control'
+    '--leo-menu-control-min-width': {
+      control: 'text',
+      description: 'The user provided min width for the menu. The menu may be wider than this value depending on its contents'
```
AlanBreck, line 21: Can we add this to document the new read only prop? I use this computed property value in premium.

```suggestion
    '--leo-menu-control-min-width': {
      control: 'text',
      description: 'The user provided min width for the menu. The menu may be wider than this value depending on its contents'
    },
    '--leo-menu-target-width': {
      description: '(readonly): Computed width of menu control'
```

AlanBreck (dismissed):

AlanBreck:
Closing as stale.

Issue #730, opened 2024-07-01
Updated focus state tokens


aguscruiz:
Based on what we've been speaking with Jay, I have an update for our focus state:
https://www.figma.com/design/g8Z0q6TMPYDq6zXh9Y7LWD?node-id=278-12159#913075082

- [ ] Changed the style from a drop shadow to a simple stroke.
- [ ] There's 2 variables of the focus style: One for `filled` and `hero` buttons, and one for everything else.

Tokens:
--leo-color-systemfeedback-focus-default: 1px outline - themable/30
--leo-color-systemfeedback-focus-button-special: 2px outline - themable/70

<img width="1355" alt="image" src="https://github.com/user-attachments/assets/1adc5eaf-5b45-40d9-bd49-a11f72aaf21e">

PR #781, opened 2024-08-16
[Button]: Update styling for various states

AlanBreck:
Resolves https://github.com/brave/leo/issues/730

### DIFF
--- a/src/components/button/button.svelte
+++ b/src/components/button/button.svelte
@@ -125,7 +125,7 @@
     // to only apply to `box-shadow` and `border-color` in .isHero
     --default-transition: box-shadow 0.12s ease-in-out, color 0.12s ease-in-out,
       border-color 0.12s ease-in-out, opacity 0.12s ease-in-out;
-    --box-shadow-hover: var(--leo-effect-elevation-02);
+    --box-shadow-hover: var(--leo-effect-elevation-01);
     --box-shadow-focus: var(--leo-effect-focus-state);
     --radius: 0;
     --border-color: transparent;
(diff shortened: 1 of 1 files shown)


aguscruiz (approved):
I think it's ok for now. Yeah, the hover on dark mode is more subtle, but I think it's acceptable and we can revisit a bit later with more time.

src/components/button/button.svelte
```diff
@@ -212,7 +213,7 @@
       color: var(--color-focus, var(--color));
       box-shadow: var(--box-shadow-focus);
       background: var(--bg-focus, var(--bg));
-      border-color: var(--border-color-focus, var(--border-color));
```
fallaciousreasoning, line 215: should we remove all the `--border-color-focus` references (I don't think we use it now)

fallaciousreasoning (approved):

src/components/button/button.svelte
AlanBreck: Good call: https://github.com/brave/leo/pull/781/commits/788ae6c87c33a77368599ea02b4b6ca14d78b04a

fallaciousreasoning (approved):
Okay, the code change LGTM, but personally I find it even harder to tell if the buttons are focused now @aguscruiz 

image

aguscruiz:
I think let's keep it for now, and I'll come back to it later with a fresher set of eyes when I have some more time. It should only be a problem on the filled button

fallaciousreasoning:
Okay, but fwiw that's the main button we use :smile:

AlanBreck:
I agree that the current focus state isn't really noticeable, but it's also not worse than what we have right now in https://nala.bravesoftware.com/?path=/story/components-button--all&globals=theme:light

Screenshot 2024-08-21 at 4 49 51 PM

aguscruiz (approved):
Approved!

PR #786, opened 2024-08-20
WIP: Add optional delay to alert center actions

AlanBreck:
https://github.com/user-attachments/assets/56c5bb1d-8966-4344-bba2-48c84413ba88

### DIFF
--- a/src/components/alert/alert.stories.svelte
+++ b/src/components/alert/alert.stories.svelte
@@ -2,17 +2,18 @@
   import { Meta, Story, Template } from '@storybook/addon-svelte-csf'
   import Button from '../button/button.svelte'
 
-  import Alert, { modes, types } from './alert.svelte'
-  import { showAlert } from './alertCenter.svelte'
-  import Icon from '../icon/icon.svelte'
   import Slot from '../../storyHelpers/Slot.svelte'
   import SlotInfo from '../../storyHelpers/SlotInfo.svelte'
+  import Icon from '../icon/icon.svelte'
+  import Alert, { modes, types } from './alert.svelte'
+  import { showAlert } from './alertCenter.svelte'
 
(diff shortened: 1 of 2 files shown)


fallaciousreasoning (commented):
Hey, so I think this is a great idea - I wonder if we can come up with a more generic approach that'd let us change buttons on the fly, rather than needing to hardcode support in the alert center.

That said, maybe this is the only thing we'd want to change?

AlanBreck:
> Hey, so I think this is a great idea - I wonder if we can come up with a more generic approach that'd let us change buttons on the fly, rather than needing to hardcode support in the alert center.
> 
> 
> 
> That said, maybe this is the only thing we'd want to change?

I suppose we could just leverage the ability to pass a custom Alert component, and leave this behavior up to consuming apps.

AlanBreck:
> a more generic approach that'd let us change buttons on the fly

Actually, mind fleshing this out a bit more?

fallaciousreasoning:
I always struggle a bit with stuff like this in Svelte:

Something like https://github.com/brave/leo/pull/786 is what I was thinking. What are your thoughts?

AlanBreck:
https://github.com/brave/leo/pull/789 is definitely a lot cleaner. I suppose that solution would still allow for updating the text, yes? The downside is that we couldn't have the ProgressRing. Unless we passed a custom Button component? Maybe we just don't support that and rely on text.

So the essential difference here is that the AlertCenter provides an API to imperatively update the alert, and the the consumer would own the logic for what constitutes an update, etc. Yes?

fallaciousreasoning:
Yeah exactly :smile: We could even add in an optional prop for the progress, I guess?

AlanBreck:
Closing in favor of https://github.com/brave/leo/pull/789

PR #789, opened 2024-08-21
[AlertCenter]: Alternate custom alerts proposal

fallaciousreasoning:


src/components/alert/alertCenter.svelte
```diff
@@ -109,11 +121,15 @@
...
-            <Button
+            <svelte:component
+              this={action.component || Button}
               size={alert.mode === "full" ? "medium" : "small"}
               fab={action.icon && !action.text}
               kind={action.kind || 'filled'}
+              isDisabled={action.isDisabled}
+              isLoading={action.isLoading}
               onClick={() => action.action(alert)}
+              {...action}
```
AlanBreck, line 132: @fallaciousreasoning, I'm getting TS errors here that I've not figured out. Haven't gotten to look too deeply yet, but if you get a moment when you come back online, I'd love your eyes on this.

src/components/alert/alert.stories.svelte
```diff
@@ -176,12 +169,10 @@
     Has action
     <input type="checkbox" bind:checked={hasAction} />
   </label>
-  {#if hasAction}
-    <label>
-      Action Load For
-      <input type="text" bind:value={actionLoadFor} />
-    </label>
-  {/if}
+  <label>
```
fallaciousreasoning, line 172: I think this should still only show up if `haveAction` is true

src/components/alert/alertCenter.svelte
```diff
@@ -120,13 +121,15 @@
...
-            <Button
+            <svelte:component
+              this={action.component || Button}
               size={alert.mode === "full" ? "medium" : "small"}
               fab={action.icon && !action.text}
               kind={action.kind || 'filled'}
               isDisabled={action.isDisabled}
               isLoading={action.isLoading}
               onClick={() => action.action(alert)}
+              {...action}
```
fallaciousreasoning, line 132: Possibly, this should go first and we could remove the `isDisabled` and `isLoading` props?

src/components/alert/alertCenter.svelte
fallaciousreasoning: I'd probably just chuck an `as any` at the end - typing it properly probably isn't worth your time here

src/components/alert/alertCenter.svelte
AlanBreck: Sure. For my own learning, though, why?

fallaciousreasoning: 'cause it's being passed in via the spread anyway

we're doing 

```
const action = { isDisabled: false, isLoading: true }
const props = { isDisabled: action.isDisabled, isLoading: action.isLoading, otherProps, ...action }

// when we could do
const props = { ...action, otherProps }
```

fallaciousreasoning:
(also, I can't :+1: this PR 'cause I created it, but LGTM)

Feel free to add a :+1: and merge.

src/components/alert/alertCenter.svelte
```diff
@@ -53,19 +63,28 @@
     }
 
     resumeDismiss() {
-      if (!this.duration) return
+      if (!this.duration || this.actions.length) return
```
AlanBreck, line 66: @fallaciousreasoning, do you agree with this?

fallaciousreasoning (commented):
LGTM (you'll have to approve though)

src/components/alert/alertCenter.svelte
fallaciousreasoning: Hmmm, I don't think so - if you set a duration it should hide then - you can set a zero duration with actions.

Probably doesn't work in the story because the duration is a string not a number. Does 

```suggestion
      if (!+this.duration) return
```

work?

AlanBreck: Sorry, the original worked, I was just thinking that it doesn't make much sense to have a timeout if there's an action.

fallaciousreasoning: I think it still does - for example, on Android we show an Alert with an `Undo` action after closing a tab for 3 seconds.

AlanBreck (approved):

PR #792, opened 2024-08-22
[SegmentedControl]: Add var to control max width

AlanBreck:
...additionally ensures that control items will be evenly split across
the entire width.

### DIFF
--- a/src/components/segmentedControl/segmentedControl.svelte
+++ b/src/components/segmentedControl/segmentedControl.svelte
@@ -121,8 +121,9 @@
     --leo-control-item-icon-gap: var(--leo-spacing-m);
     --leo-control-item-font: var(--leo-font-components-button-default);
 
-    display: flex;
-    max-width: max-content;
+    display: grid;
+    grid-auto-flow: column;
+    max-width: var(--leo-segmented-control-max-w, max-content);
     align-items: center;
     position: relative;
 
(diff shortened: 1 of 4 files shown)


fallaciousreasoning (approved):

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -121,8 +121,9 @@
     --leo-control-item-icon-gap: var(--leo-spacing-m);
     --leo-control-item-font: var(--leo-font-components-button-default);
 
-    display: flex;
-    max-width: max-content;
+    display: grid;
+    grid-auto-flow: column;
+    max-width: var(--leo-segmented-control-max-w, max-content);
```
AlanBreck, line 126: @fallaciousreasoning and @petemill, or do we want to simply remove the `max-content` altogether?

fallaciousreasoning: I think it's probably safe to remove?

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -122,8 +122,7 @@
     --leo-control-item-font: var(--leo-font-components-button-default);
 
     display: flex;
-    min-width: max-content;
-    max-width: var(--leo-segmented-control-max-w, none);
+    width: var(--leo-segmented-control-width, fit-content);
```
fallaciousreasoning, line 125: So, just to make things much more tricky for you could you make this work with `:host`?

Maybe something like

```css

:host {
    width: var(--leo-segmented-control-width, fit-content);

    & .leo-segmented-control {
        width: 100%;
    }
}
```

and test it works in the plain-html examples

fallaciousreasoning (approved):

PR #793, opened 2024-08-25
[Tokens]: Update universal variables

fallaciousreasoning:


src/tokens/universal.variables.json
```diff
@@ -3791,14 +3781,19 @@
...
           },
           "background-selected-item": {
             "type": "color",
             "value": "#f0f0f4",
             "referencedVariable": "$primitive.neutral.95"
           },
+          "background-hover-item": {
+            "type": "color",
+            "value": "#ffffff",
+            "referencedVariable": "$primitive.neutral.100"
```
fallaciousreasoning, line 3795: Hey, this doesn't look great in Light Mode:
image

(dark mode looks fine)

fallaciousreasoning: (`Hello World` is being hovered)

fallaciousreasoning: I suspect

Active (this is just the background for the open omnibar): 100
Hovered: 98
Selected: 95

fallaciousreasoning: FWIW, I'm looking at the code and it looks like it uses the following logic for the Omnibox:

1. Get the max contrasting color (white in dark, black in light) and mix:
2. 10% of that into the background for hover
3. 16% of that into the background for selected

I think we should keep this logic, because it means we can handle high contrast mode too.

aguscruiz: Yeah let's keep that logic, I'm fine with that 👍 
If you find more of these cases let me know, I I'm doing my color matching based on screenshots and figuring out my token that is closes to that contrast. 

But if they're doing calculations for theirs, then we should probably keep their logic.

fallaciousreasoning:
Closing, as we merged @aguscruiz's PR :tada:

PR #796, opened 2024-08-27
[Tailwind-email]: Add new config for tailwind-email

AlanBreck:
This is a greatly pared down version of the Tailwind config with
the constraints of HTML email development in mind.


fallaciousreasoning (commented):
Hey, this is just a more general comment before I do a more in depth review, but do you think Tailwind is the right approach for email?

As far as I knew, they rely pretty heavily on variables in their internals (for example, handling opacity), and that sort of stuff probably won't work properly in emails?

(I might just be imagining that this is how it works though, and if this works fine for email I guess we're all good?)

AlanBreck:
@fallaciousreasoning, normally yes, however Maizzle changes that. The immediate use case is here: https://github.com/brave-experiments/email-template-builder

Maybe we should name it `tailwind-maizzle` to make that dependency more clear? It doesn't _have_ to be used with Maizzle, strictly speaking, but Maizzle handles all of the inlining, variable resolution, etc.

fallaciousreasoning:
Cool, good to know - I'll have a proper look over this tomorrow :smile:

src/tokens/transformation/tailwind-email/static/index.js
fallaciousreasoning, line 1: can this be a typescript file?

src/tokens/transformation/tailwind-email/static/package.json
```diff
@@ -0,0 +1,3 @@
+{
```
fallaciousreasoning, line 1: isn't the parent directory already commonjs?

src/tokens/transformation/tailwind-email/copyStaticFiles.js
```diff
@@ -0,0 +1,23 @@
+const { join } = require('path')
+const { readdirSync, unlink, cpSync, rmdir, statSync } = require('fs')
+
+const staticFilesPath = join(__dirname, './static')
+const staticFiles = readdirSync(staticFilesPath)
+
+module.exports = {
+  do: function (dictionary, config) {
+    const targetDir = join(config.buildPath, config.preset)
```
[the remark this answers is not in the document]
fallaciousreasoning: this is fine.

src/tokens/transformation/tailwind-email/copyStaticFiles.js
```diff
@@ -0,0 +1,23 @@
+const { join } = require('path')
+const { readdirSync, unlink, cpSync, rmdir, statSync } = require('fs')
+
+const staticFilesPath = join(__dirname, './static')
+const staticFiles = readdirSync(staticFilesPath)
+
+module.exports = {
+  do: function (dictionary, config) {
```
fallaciousreasoning, line 8: nit: 
```suggestion
  do: (dictionary, config) => {
```

or


```suggestion
  do(dictionary, config) {
```

src/tokens/transformation/tailwind-email/copyStaticFiles.js
fallaciousreasoning, line 1: Can this file be typescript?

src/tokens/transformation/tailwind-email/formatTokens.ts
```diff
@@ -0,0 +1,159 @@
+import merge from 'lodash.merge'
+import { Formatter } from 'style-dictionary'
+
+const themes = ['light', 'dark']
+
+const kebabCase = (str: string) => str && str.toLowerCase().replaceAll(' ', '-')
```
fallaciousreasoning, line 6: I'm pretty sure we have this function (or an equivalent) somewhere else, right? Could we share it (maybe in a casing helpers file)?

src/tokens/transformation/tailwind-email/formatTokens.ts
```diff
@@ -0,0 +1,159 @@
...
+ * structure ready for Tailwind. The conditional statements
+ * are largely to handle creating both static and dynamic
+ * tokens. E.g. A given token needs three variants:
+ * dynamic, static light, and static dark. This function gets
+ * run twice: once to create all static tokens, and another
+ * time to create the dynamic tokens which places values on
+ * the parent (e.g. the root object, or 'legacy') and uses the
+ * appropriate color variable.
+ */
+function createColorTokensFromGroup(tokens, staticTheme = true) {
```
fallaciousreasoning, line 19: Okay, so the main different here is we output light/dark variables, unlike in normal Tailwind where we can update the colors dynamically?

src/tokens/transformation/tailwind-email/formatTokens.ts
```diff
@@ -0,0 +1,159 @@
...
+  const colorTokens = {}
+  tokens.forEach(({ type, name, ...t }) => {
```
fallaciousreasoning, line 21: optional nit: I think its generally nicer/more readable/faster to use a `for .. of` loop

src/tokens/transformation/tailwind-email/formatTokens.ts
fallaciousreasoning: Also, any chance we could throw a type on `tokens`?

src/tokens/transformation/tailwind-email/formatTokens.ts
```diff
@@ -0,0 +1,159 @@
...
+
+      if (Object.keys(colorGroup).length > 0) {
+        colorTokens[t.attributes.type] = colorGroup
+      }
+    }
+  })
+  return colorTokens
+}
+
+export default (({ dictionary }) => {
```
fallaciousreasoning, line 102: nit: Can we throw a type on dictionary?


```suggestion
export default (({ dictionary }: { dictionary: Dictionary }) => {
```

src/tokens/transformation/tailwind-email/formatTokens.ts
```diff
@@ -0,0 +1,159 @@
...
+  const borderRadii = new Map([['none', '0']])
+  const spacing = new Map<string | number, string | number>([[0, 0]]) // Initialize with option for 0 spacing
+  const gradients = new Map()
+  const boxShadows = new Map([['none', 'none']])
+  const dropShadows = new Map<string, string | string[]>([
+    ['none', '0 0 #0000']
+  ])
+
+  // Format all other tokens
+  dictionary.allTokens.forEach(({ type, name, ...t }) => {
```
fallaciousreasoning, line 114: optional nit: `for .. of`

src/tokens/transformation/tailwind-email/index.ts
fallaciousreasoning, line 1: hey @AlanBreckthis looks pretty similar to the one in `tailwind` any chance we could just reuse that one?

fallaciousreasoning: Maybe as 

```ts
import '../tailwind'

// do customizations
```

fallaciousreasoning (commented):
Hey, so this seems fine - my only concern is that there's definitely a bit of duplication between this and the `tailwind` transformation, and I worry a bit about falling out of sync.

WDYT? It'd be great if we could share stuff (where it makes sense), as these two targets are kinda similar.

PR #814, opened 2024-09-04
[Button]: prevent consumer from overriding disabled atttribute

AlanBreck:
With the current configuration, a consumer can disable a button via either `<Button isDisabled={true}>` or `<Button disabled={true}>`. However, these can be in conflict, and if the consumer accidentally does something like `<Button isDisabled={true} disabled={false}>`, then the `disabled` value wins.

A more likely conflict is if the user sets `isLoading` to true, but has a different condition for `disabled` which is `false`:

```svelte
<Button isLoading={isFetching} disabled={shouldDisableButtons}>
```

This PR attempts to avoid this by not allowing the consumer provided `disabled` attribute to override the `disabled` state set by the `Button` component.

### DIFF
--- a/src/components/button/button.svelte
+++ b/src/components/button/button.svelte
@@ -70,9 +70,9 @@
   class:isTiny={size === 'tiny'}
   class:fab
   class:isLoading
-  disabled={isLoading || disabled || undefined}
   on:click={onClick}
   {...$$restProps}
+  disabled={isLoading || disabled || undefined}
 >
   {#if isLoading}
     <div class:content={!fab}>


AlanBreck:
If we had some way of deprecating properties, we could do that now, right?

fallaciousreasoning:
As a comment on the export, maybe?

```svelte
/**
 * Deprecated - use disabled instead
 * /
export let isDisabled = false;
```

AlanBreck:
Unfortunately, there's no great path for deprecation right now apart from a console log if it's detected, but a moderator from the Svelte Discord channel kindly created an issue on `sveltejs/language-tools` requesting support for `@deprecated`: https://github.com/sveltejs/language-tools/issues/2491

fallaciousreasoning (approved):

PR #837, opened 2024-09-18
[Renovate]: Enable automerge for devDependencies, minor and patches

fallaciousreasoning:
I was talking to @thypon about disabling dependabot updates for `/examples/example-ui-react` or setting them to automerge and thought, why not set it for a bunch of things we normally merge anyway. If anything breaks, the build in Nala should fail, which will prevent the automerge.

Interested in hearing some thoughts on this (and whether its a bad idea).

For context, this was my original message to @thypon:

> Hey, I have a question - we have an example project in https://github.com/brave/leo/tree/main/examples/example-ui-react. Its not published/deployed anywhere and is only really used as a test bed during development to see if the React bindings are working properly.
Would it be safe to reduce the frequency of the dependabot/rennovate PRs (or set up automerge in this folder) as there are a lot of them, all the time, and they can cover up the important updates for the main repository.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -13,6 +13,11 @@
       "automerge": true
     },
     {
+      "matchUpdateTypes": ["digest"],
+      "matchRepositories": ["^brave/.*$"],
+      "automerge": true
+    },
+    {
       "matchDepTypes": ["devDependencies"],
       "automerge": true
     }
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["local>brave/renovate-config"],
   "labels": ["dependencies", "renovate"],
   "prConcurrentLimit": 4,
@@ -10,6 +11,7 @@
     },
     {
       "matchUpdateTypes": ["minor", "patch", "pin"],
+      "minimumReleaseAge": "7 days",
       "automerge": true
     },
     {
@@ -19,6 +21,13 @@
     },
     {
       "matchDepTypes": ["devDependencies"],
+      "minimumReleaseAge": "7 days",
+      "automerge": true
+    },
+    {
+      "matchManagers": ["github-actions"],
+      "enabled": true,
+      "minimumReleaseAge": "7 days",
       "automerge": true
     }
   ],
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -1,13 +1,36 @@
...
+    "dependencies",
+    "renovate"
+  ],
   "prConcurrentLimit": 4,
   "rangeStrategy": "pin",
   "packageRules": [
     {
-      "matchDepTypes": ["peerDependencies"],
+      "matchDepTypes": [
+        "peerDependencies"
```
fallaciousreasoning, line 14: if we don't disable peerDependencies, it tries to set React to latest, which isn't compatible with brave-core (most of the time)

.github/renovate.json
```diff
@@ -1,13 +1,36 @@
...
+      ],
       "enabled": false
+    },
+    {
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
```
fallaciousreasoning, line 21: I'm not opposed to adding `"digest"` and `"pin"` to this list, but I'm a bit hazy on exactly what they mean.

The build will fail if something goes wrong here.

.github/renovate.json
```diff
@@ -1,13 +1,36 @@
...
+      ],
+      "automerge": true
+    },
+    {
+      "matchDepTypes": [
+        "devDependencies"
```
fallaciousreasoning, line 27: ditto

AlanBreck (approved):

diracdeltas (changes requested):
my 2 concerns with this are making sure `npm audit` and the Socket checks are passing before the automerge happens (need to make sure those things are set up as blocking github checks in this repo and that the automerge doesn't override that); and that we set a delay of ~1 week before automerge happens so there is time to detect if these updates are malicious.

fallaciousreasoning:
@diracdeltas I've set up the `audit` check in https://github.com/brave/leo/pull/871 but as per our messages, I've disabled it for `rennovate` and `dependabot` PRs. Is that going to be a problem?

diracdeltas (approved):

fallaciousreasoning:
Ran the rennovate config validator, and it seems happy, so I'm going to merge this.
image


cc @AlanBreck if we should keep an eye on whether dependabot/rennovate PRs are still landing.

PR #858, opened 2024-10-07
Add themable "5" - Remove Poppins from buttons

aguscruiz:
- We're going to test removing Poppins from our buttons in favor of Inter.
- Additionally, we're adding a "5" shade to all our themable colors for lighter versions when needed


aguscruiz:
When you have a chance could you review this one? 🥺

fallaciousreasoning (approved):

PR #868, opened 2024-10-14
[FormItem]: Make it possible to customize border color

fallaciousreasoning:
@aguscruiz wants us to remove it on the SearchBox on the NTP

### DIFF
--- a/src/components/formItem/formItem.svelte
+++ b/src/components/formItem/formItem.svelte
@@ -27,7 +27,7 @@
       description: 'The padding of the control',
       type: 'string'
     },
-    '--leo-control-border': {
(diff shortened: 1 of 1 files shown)


src/components/formItem/formItem.svelte
```diff
@@ -27,6 +27,10 @@
       description: 'The padding of the control',
       type: 'string'
     },
+    '--leo-control-border': {
```
petemill, line 30: Is this missing the `-color` suffix?

fallaciousreasoning: why yes, yes it is :D

petemill (approved):

PR #877, opened 2024-10-20
[Tokens]: Correctly process font-weight

fallaciousreasoning:


aguscruiz (approved):
If this works as intended then great! :D

src/tokens/transformation/web/webSize.ts
```diff
@@ -2,10 +2,13 @@ import { Transform } from 'style-dictionary'
...
   type: 'value',
-  matcher: function (token) {
-    return token.type === 'number'
+  matcher({ type }) {
+    return type === 'number'
   },
-  transformer: function ({ value }) {
-    return `${value}px`
+  transformer({ value, name }) {
+    // Font weight doesn't have a dimension - unfortunately we have not better
```
AlanBreck, line 9: nit:

```suggestion
    // Font weight doesn't have a dimension - unfortunately we have no better
```

src/tokens/transformation/web/webSize.ts
```diff
@@ -2,10 +2,13 @@ import { Transform } from 'style-dictionary'
...
+    // way of determining whether this is a font-weight than checking the name.
+    const dimension = name.endsWith('font-weight') ? '' : 'px'
+    return `${value}${dimension}`
```
AlanBreck, line 12: Should this be `unit` instead of `dimension`?

PR #878, opened 2024-10-21
[Web Components]: Recreate component on connect/disconnect.

fallaciousreasoning:


fallaciousreasoning:
@szilardszaloki I tested with your minimal repro and this seems to fix it. Mind installing it into your branch and seeing if it helps?

(you might need to hack a couple of C++ errors about tokens, but don't worry about a proper fix - I can bump Nala separately once we verify this fixes your issue).

szilardszaloki:
@fallaciousreasoning Yup, no flickering. 👍🏻

petemill (approved):

PR #884, opened 2024-10-27
[WebComponents]: Ignore pure whitespace nodes

fallaciousreasoning:
@szilardszaloki has been migrating some components to LitElement and noticed that the `leo-icon` elements weren't working.

Our working theory is that `Lit` injects whitespace into its HTML elements (there's a slotted `\n`):
https://search.brave.com/search?q=litelement+adds+whitespace+to+webcomponents&source=desktop&summary=1&summary_og=f0f84d159a7ddd8a5ff175

### DIFF
--- a/src/components/svelte-web.ts
+++ b/src/components/svelte-web.ts
@@ -134,8 +134,11 @@ export default function registerWebComponent(
       const slotsNames = Array.from(this.children).map((c) =>
         c.getAttribute('slot')
       )
-      // Add default slot if there are nodes without a slot name.
-      if (this.childNodes.length > slotsNames.length) slotsNames.push(null)
+      // Add default slot if there are non-empty nodes without a slot name.
+      const nonEmptyNodes = Array.from(this.childNodes).filter(
+        (c) => c.nodeName !== '#text' || c.textContent.trim().length
+      )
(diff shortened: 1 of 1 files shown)


szilardszaloki (approved):

AlanBreck (approved):

src/components/svelte-web.ts
```diff
@@ -134,8 +134,11 @@ export default function registerWebComponent(
       const slotsNames = Array.from(this.children).map((c) =>
         c.getAttribute('slot')
       )
-      // Add default slot if there are nodes without a slot name.
-      if (this.childNodes.length > slotsNames.length) slotsNames.push(null)
+      // Add default slot if there are non-empty nodes without a slot name.
+      const nonEmptyNodes = Array.from(this.childNodes).filter(
```
petemill, line 138: `this.childNodes` will always be nodes without a slot name?

fallaciousreasoning: Nah, `childNodes` includes all childNodes of the element (including `text` nodes, which can't have a slot attribute).

This is basically to handle the case where the slotted content is `#text` (which won't show up in the `children` array) but which indicates we have a default slot.

This was working well before, but Lit is slotting whitespace into the element which means we're adding a default slot to our component (which in this case, swaps the icon to child element mode, and doesn't render the named icon).

PR #890, opened 2024-10-30
[Svelte]: Update svelte-web to work with Svelte5

fallaciousreasoning:
Minimal example of the changes we need to support Svelte5  in our web components wrapper

@AlanBreck we probably shouldn't land this until the projects we have which use Svelte are ready for this :fearful: 

**Note:** At this point Storybook doesn't support Svelte5 either


src/tokens/transformation/tailwind/extractComponentStyles.js
```diff
@@ -52,6 +52,9 @@ module.exports = {
           }
         )
 
+        // TODO: Beg Jacob for help
+        return;
+
```
fallaciousreasoning, line 57: halp @AlanBreck!

rollup.config.js
```diff
@@ -66,7 +66,9 @@ export default {
       // Don't emit CSS - it doesn't work properly with Web Components.
       emitCss: false,
       compilerOptions: {
-        customElement: false
+        customElement: false,
+        css: 'injected',
+        accessors: true,
```
fallaciousreasoning, line 71: There's a warning that `accessors: true` isn't supported in runes mode, which seems odd

PR #899, opened 2024-11-12
[Dialog]: hide actions if slot is empty

AlanBreck:


fallaciousreasoning (approved):
lgtm assuming it works for WebComponents too

fallaciousreasoning (approved):

PR #912, opened 2024-11-22
**Breaking** [Alert] Introduce notice type, remove mode prop, and title slot will always show if it has content

petemill:
If a title slot was provided and the mode wasn't set to full then the title would never show. Seems to me that the full mode is only for adjustments made to make the alert look good with a title - ~so this PR forces 'full' mode when a title is present (or hasTitle prop is true for when the framework can't have an optional slot!)~. After discussion with @aguscruiz this PR removes `mode="full"` (leaving only a single mode, and removing it) as it wasn't being used by any designs, and allows title without any mode restriction

This also adds a variant for 'notice' type, which is color-less and has smaller features.

image
image

### DIFF
--- a/src/components/alert/alert.svelte
+++ b/src/components/alert/alert.svelte
@@ -4,9 +4,6 @@
   export const types = ['info', 'warning', 'error', 'success', 'notice'] as const
   export type AlertType = (typeof types)[number]
 
-  export const modes = ['simple', 'full'] as const
-  export type AlertMode = (typeof modes)[number]
-
   const defaultIcons: { [P in AlertType]: IconName } = {
     'info': 'info-filled',
     'error': 'warning-circle-filled',
@@ -20,22 +17,19 @@
   import Icon from '../icon/icon.svelte'
 
   export let type: AlertType = 'error'
-  export let mode: AlertMode = 'simple'
   export let isToast = false
-  export let hasActions = $$slots.actions
-  export let hasContentAfter = $$slots['content-after']
-  export let hasTitle = $$slots.title
   export let hideIcon = false
 
+  // TODO: Remove when only supporting svelte >5 which can render slotted content conditionally
+  export let hasActions = $$slots.actions
+  export let hasContentAfter = $$slots['content-after']
+
   $: currentType = type ?? 'error'
-  $: currentMode = hasTitle ? 'full' : mode ?? 'simple'
 </script>
 
 <div
   class="leo-alert {currentType}"
   class:toast={isToast}
-  class:simple={currentMode === 'simple'}
-  class:full={currentMode === 'full'}
 >
   {#if !hideIcon}
   <div class="icon">
@@ -45,7 +39,7 @@
   </div>
   {/if}
   <div class="content">
-    {#if hasTitle && $$slots.title}
+    {#if $$slots.title}
       <div class="title">
         <slot name="title" />
       </div>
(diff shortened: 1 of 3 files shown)


src/components/alert/alert.svelte
```diff
@@ -21,20 +24,18 @@
   export let isToast = false
   export let hasActions = $$slots.actions
   export let hasContentAfter = $$slots['content-after']
+  export let hasTitle = $$slots.title
```
fallaciousreasoning, line 27: Hey @petemill what's the point of exporting this? I feel like it also might be slightly confusing to set `mode=simple` and provide a title and have the alert show in `full` mode?

I think at least we shouldn't export this :smile:

Additionally, I think `showTitle` is probably a better name, as we hide the title when this is false (and we might have a title, and just decide not to show it)

src/components/alert/alert.svelte
```diff
@@ -44,7 +45,7 @@
   </div>
   {/if}
   <div class="content">
-    {#if mode === 'full' && $$slots.title}
+    {#if hasTitle && $$slots.title}
```
fallaciousreasoning, line 48: Are we doing this for Svelte, where you can't conditionally render a slot?

AlanBreck: Guessing yes, and that it's the same reason for [exporting `hasTitle`](https://github.com/brave/leo/pull/912#discussion_r1855581102).

fallaciousreasoning: Are there cases where we're going to conditionally show a title though? I feel like that's generally decided upfront

src/components/alert/alert.svelte
petemill: Well yes I was confused as to why we are exporting `hasActions` and `hasContentAfter` and guessed it was for svelte (and storybook), in order to hide the title element when no title is provided.

fallaciousreasoning (approved):
Happy to land this as-is 'cause its the same pattern as everything else :shrug:

src/components/alert/alert.svelte
AlanBreck: > Are there cases where we're going to conditionally show a title though? I feel like that's generally decided upfront

Not that I'm aware of for `title`.

petemill: I see what you're saying - the svelte issue is about having slotted content in a templated conditional. In any case, I just tested what happens when we render the empty title element and there are no visual issues, so I might as well let it get rendered...

fallaciousreasoning (approved):
Can you mark this PR as **BREAKING** so we know in the changelog?

src/components/alert/alert.svelte
```diff
@@ -1,40 +1,35 @@
...
+    'success': 'check-circle-filled',
+    'notice': 'info-outline'
   }
 </script>
 
 <script lang="ts">
   import Icon from '../icon/icon.svelte'
 
   export let type: AlertType = 'error'
-  export let mode: AlertMode = 'simple'
```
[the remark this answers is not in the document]
petemill: there is no concept of `full` - there is no longer an icon size difference

AlanBreck:
> Can you mark this PR as BREAKING so we know in the changelog?

What about it will be breaking?

petemill:
> > Can you mark this PR as BREAKING so we know in the changelog?
> 
> What about it will be breaking?

If any Alerts were using `mode="full"` it will cause an error?

PR #914, opened 2024-11-26
[Dialog]: **BREAKING** Replace 'size' property with media query

AlanBreck:
NOTE: this purpose of this PR is _not_ to bring the Dialog component up to date with the latest in Figma, but rather to provide consumers a way of automatically switching between `mobile` and `normal` sizes via screen size.

- **[Dialog]: add customizable media queries for switching to mobile styles**
- **[Dialog]: change position of action buttons right aligned**
- **[Dialog]: scroll entire dialog instead of just .body**

### DIFF
--- a/src/components/dialog/dialog.svelte
+++ b/src/components/dialog/dialog.svelte
@@ -154,7 +154,7 @@
   }
 
   .leo-dialog.hasHeader {
-    grid-template-rows: auto 1fr;
+    grid-template-rows: auto auto;
   }
 
   /** Since Svelte 4 doesn't support conditional slots in the consumer,
@@ -163,12 +163,12 @@
    * case the selector with :host. */
   :host .leo-dialog.hasActions,
   .leo-dialog.hasActions:has([slot='actions']:not(:empty)) {
-    grid-template-rows: 1fr auto;
+    grid-template-rows: auto auto;
   }
 
   :host .leo-dialog.hasHeader.hasActions,
   .leo-dialog.hasHeader.hasActions:has(.actions [slot='actions']:not(:empty)) {
-    grid-template-rows: auto 1fr auto;
+    grid-template-rows: auto auto auto;
   }
 
   .leo-dialog:not(.modal) {
@@ -209,7 +209,6 @@
   }
 
   .leo-dialog .body {
-    overflow-y: auto;
     background: var(--background);
     color: var(--leo-color-text-secondary);
     font: var(--leo-font-default-regular);
--- a/src/components/dialog/dialog.svelte
+++ b/src/components/dialog/dialog.svelte
@@ -261,7 +261,7 @@
     display: flex;
     flex-direction: row;
     align-items: center;
-    justify-content: stretch;
+    justify-content: end;
     gap: var(--leo-spacing-xl);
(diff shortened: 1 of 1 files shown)


src/components/dialog/dialog.svelte
```diff
@@ -42,6 +45,20 @@
     isOpen = false
     onClose?.()
   }
+
+  const setSize = (matches: boolean) => {
+    size = matches ? 'mobile' : 'normal'
+  }
+
+  onMount(() => {
```
fallaciousreasoning, line 53: Maybe this should be a reactive statement so the media query updates when `mobileBreakpoint` is changed

src/components/dialog/dialog.svelte
```diff
@@ -42,6 +45,20 @@
     isOpen = false
     onClose?.()
   }
+
+  const setSize = (matches: boolean) => {
```
fallaciousreasoning, line 49: this could be scoped to `onMount`

fallaciousreasoning (dismissed):
lgtm with comments

src/components/dialog/dialog.svelte
AlanBreck: Not opposed, but what's the advantage in this case?

fallaciousreasoning: Just keeps it defined where its used :smile:

src/components/dialog/dialog.svelte
```diff
@@ -23,10 +24,11 @@
   export let modal = true
   export let showClose = false
   export let showBack = false
-  export let size: DialogSizes = 'normal'
+  export let size: DialogSizes = undefined
   export let escapeCloses = true
   export let backdropClickCloses = true
   export let animate = true
+  export let mobileBreakpoint = 375
```
petemill, line 31: Should this be specified via css?

AlanBreck: The idea was to allow consumers to determine the breakpoint so that it's not locked in.

fallaciousreasoning (approved):

src/components/dialog/dialog.svelte
```diff
@@ -141,8 +136,8 @@
     border: none;
     display: grid;
 
-    width: var(--leo-dialog-width, 500px);
-    max-width: calc(100% - var(--leo-spacing-m) * 2);
+    width: calc(100% - var(--leo-spacing-m) * 2);
```
fallaciousreasoning, line 139: does it work without this calc? I feel like it should because the padding is applied inside the dialog element (i.e. this will result in a `4xl` padding around the dialog, rather than a 2xl)

fallaciousreasoning (approved):
lgtm with one comment

src/components/dialog/dialog.svelte
AlanBreck: The `calc` isn't for the padding, but rather the margin between the viewport and the dialog.

| Without `calc` | With `calc` |
| -- | -- |
| SCR-20241206-senr | SCR-20241206-sfbj |

petemill (approved):
It's a good direction to go in - let's see how it goes!

PR #915, opened 2024-11-26
Update to toggle style

aguscruiz:
Updated the toggle design to match Figma.


src/components/toggle/toggle.svelte
```diff
@@ -241,6 +241,7 @@
       .thumb {
         --thumb-offset: var(--checked-thumb-offset);
         color: var(--checked-color);
+        --thumb-disabled-color: var(--leo-color-schemes-on-primary);
```
AlanBreck, line 244: This breaks the themeability of the toggle:
SCR-20241126-jnjx

aguscruiz: Ugh :(

aguscruiz: @fallaciousreasoning @AlanBreck  shouldn't `--leo-color-schemes-on-primary` change depending on the color that is chosen as primary? Or how should I do that? 


Btw I just realized this is also broken on the working version? If I customize the color there it doesn't show it on the thumb
https://nala.bravesoftware.com/?path=/docs/components-toggle--docs&globals=theme:dark

aguscruiz: Ok, I think I fixed it now.

I can change the color to whatever I want and it doesn't break on the default colors

image

image

AlanBreck (approved):

Issue #574, opened 2024-03-01
[Update] Label change for default background color


aguscruiz:
I hada to make a small change to labels, they're:

- Changed background color of default ones, from using themable-20, to themable-10

I already pushed the style change in my branch

![image](https://github.com/brave/leo/assets/15734630/4fa4e64a-3a7f-4d35-bf08-9383f1feda12)

Updated file here:
https://www.figma.com/file/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?type=design&node-id=195-15172&mode=design&t=LZeAk6OniOhmfMGx-11

PR #916, opened 2024-11-26
Update label style to match figma

aguscruiz:
Resolves https://github.com/brave/leo/issues/574

Update label styles to match Figma's latest version.
https://www.figma.com/design/g8Z0q6TMPYDq6zXh9Y7LWD/%F0%9F%8C%8E-Universal?m=auto&node-id=195-15172&t=YtEkhqaM1E4QR5ZC-1

image

### DIFF
--- a/src/components/label/label.svelte
+++ b/src/components/label/label.svelte
@@ -23,16 +23,16 @@
 
   $: background =
     mode === 'default'
-      ? `var(--leo-color-${color}-20)`
+      ? `var(--leo-color-${color}-10)`
       : mode === 'loud'
-        ? `var(--leo-color-${color}-50)`
+        ? `var(--leo-color-${color}-40)`
         : 'transparent'
 
   $: text =
     mode === 'default'
-      ? `var(--leo-color-${color}-50)`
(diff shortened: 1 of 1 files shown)


AlanBreck (approved):

src/components/label/label.svelte
```diff
@@ -23,16 +23,16 @@
...
+        ? `var(--leo-color-${color}-40)`
         : 'transparent'
 
   $: text =
     mode === 'default'
-      ? `var(--leo-color-${color}-50)`
+      ? `var(--leo-color-${color}-60)`
       : mode === 'loud'
-        ? `var(--leo-color-${color}-10)`
+        ? `var(--leo-color-white)`
```
fallaciousreasoning, line 35: this will apply in light & dark mode - is that intentional?

fallaciousreasoning (approved):

src/components/label/label.svelte
aguscruiz: Yup, it's intended that way, it's a bit more like iOS in that way

Issue #932, opened 2024-12-09
`undefined` string is shown on the credential loaded informative message


GeetaSarvadnya:
`undefined` string is shown on the credential loaded informative message

![Image](https://github.com/user-attachments/assets/d964b1b6-eea3-45ea-89d5-9ae9026bcdca)

PR #933, opened 2024-12-09
[AlertCenter]: fixes 'undefined' title when no title provided

AlanBreck:
Resolves https://github.com/brave/leo/issues/932

### DIFF
--- a/src/components/alert/alert.stories.svelte
+++ b/src/components/alert/alert.stories.svelte
@@ -20,7 +20,7 @@
     showAlert(
       {
         content,
-        title,
+        title: title || undefined,
         type: type ?? 'error',
         actions: hasAction
           ? [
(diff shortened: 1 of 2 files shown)


src/components/alert/alert.stories.svelte
```diff
@@ -20,7 +20,7 @@
     showAlert(
       {
         content,
-        title,
+        title: title || undefined,
```
AlanBreck, line 23: This just makes it easier to verify the behavior in Storybook since it doesn't pass a value if the string is empty.

src/components/alert/alertCenter.svelte
```diff
@@ -114,7 +114,7 @@
     >
       <svelte:component this={alert.component || Alert} {...alert} hasActions={alert.actions.length > 0} hasContentAfter={alert.canDismiss} isToast>
         <div slot="title">
-          {alert.title}
+          {alert.title ?? ""}
```
fallaciousreasoning, line 117: Wow, super weird it renders undefined here instead of nothing :fearful:

fallaciousreasoning (approved):

PR #940, opened 2024-12-13
[ButtonMenu] onclick does need stopPropagation after all to stop the Menu being blurred and closing itself

petemill:
But we can also preventDefault to stop any parent default behavior we were attempting to fix


petemill:
I'm merging this because it's a partial revert of https://github.com/brave/leo/pull/938 which has broken any use of ButtonMenu

PR #941, opened 2024-12-14
Avoid deprecated meta-tag pattern

jonathansampson:
Storybook now expects an exported meta object at the module level, rather than a <Meta /> element. This change modifies the stories to adhere to the new pattern. In a couple cases the `name` attribute of a `<Story />` element needed to be changed to avoid attempted reuse of an identifier within the same context.

### DIFF
--- a/src/components/alert/alert.stories.svelte
+++ b/src/components/alert/alert.stories.svelte
@@ -1,8 +1,37 @@
+<script context="module">
+  import Alert, { types } from './alert.svelte'
+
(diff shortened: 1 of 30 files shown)


fallaciousreasoning (approved):
Storybook drives me mental - why do they always change the way you're meant to do things? :laughing: 

Thanks for making this change @jonathansampson! I did it in brave-core and it was a super manual process, so really appreciate it!

PR #944, opened 2024-12-17

brave-support-admin:
> tsx ./src/scripts/update-icons.ts

- has a raster image element
- has a raster image element
- has a raster image element


icons/premium-indicator-color.svg
AlanBreck, line 1: @aguscruiz, will Search be moving to this icon as well?

aguscruiz: Search isn't really using the other one either, so it shouldn't be a problem. At least not for now until we really define it

AlanBreck (approved):

fallaciousreasoning (approved):

fallaciousreasoning (approved):

PR #995, opened 2025-02-19
[Deps]: Override to latest version of serialize-js (no patched version of rollup-plugin-terser)

fallaciousreasoning:


package.json
```diff
@@ -44,16 +44,16 @@
...
     "postcss-sort-media-queries": "4.4.1",
     "prettier": "3.4.2",
     "prettier-plugin-svelte": "3.3.3",
     "rollup": "3.29.5",
     "rollup-plugin-svelte": "7.2.2",
-    "sass": "1.84.0",
+    "sass": "1.83.4",
     "sass-loader": "13.3.3",
-    "storybook": "8.5.4",
+    "storybook": "8.5.3",
```
AlanBreck, line 56: Why are we reverting storybook?

package.json
```diff
@@ -5,14 +5,14 @@
   "main": "index.js",
   "type": "module",
   "dependencies": {
-    "@storybook/test": "8.5.4",
+    "@storybook/test": "8.5.3",
```
diracdeltas, line 8: did you mean to downgrade a bunch of these packages?

fallaciousreasoning:
Nice catch both of you - I think I must've made my change on one branch and then changed branch and saved the version of the file open in my editor :fearful:

diracdeltas (changes requested):
actually we don't use serialize-javascript in this package-lock.json. the dependabot alert is for [examples/example-ui-react/package-lock.json](https://github.com/brave/leo/blob/-/examples/example-ui-react/package-lock.json). we can either ignore it as a false positive (if it's just used for an example) or fix in that directory.

PR #996, opened 2025-02-21
[Dialog]: ensure dialog doesn't unnecessarily fill screen height

AlanBreck:
For some reason, the dialog viewed on ipads tries to take up the entire screen.

| before | after |
| -- | -- |
| Simulator Screenshot - iPad (10th generation) - 2025-02-21 at 16 22 16 | Simulator Screenshot - iPad (10th generation) - 2025-02-21 at 16 22 35 |

### DIFF
--- a/src/components/dialog/dialog.svelte
+++ b/src/components/dialog/dialog.svelte
@@ -135,6 +135,7 @@
     margin: auto;
     border: none;
     display: grid;
+    align-content: center;
 
     width: calc(100% - var(--leo-spacing-m) * 2);
     max-width: var(--leo-dialog-width, 374px);


petemill (approved):
Weirdly, I'm seeing a difference compared to main branch when putting test content in the dialog - the line-height is different. But it doesn't seem related.

aguscruiz:
It looks ok now on my iPad but unlike your screenshot, I seem to be getting an outline around it, maybe it's being forced into a focus state that it shouldn't have?

image

fallaciousreasoning (approved):

AlanBreck:
> I seem to be getting an outline around it, maybe it's being forced into a focus state that it shouldn't have?

Possibly. Do you only experience this on iPad? Does it also happen on nala.bravesoftware.com?

I doubt this is related to any changes in this PR, so we don't need to block.

aguscruiz:
> > I seem to be getting an outline around it, maybe it's being forced into a focus state that it shouldn't have?
> 
> Possibly. @aguscruiz, do you only experience this on iPad? Does it also happen on nala.bravesoftware.com?
> 
> I doubt this is related to any changes in this PR, so we don't need to block.

Oh yeah, also happens on nala.bravesoftware.com ✍️

PR #999, opened 2025-02-26
[Dialog]: Don't fire close event when `display: none`

fallaciousreasoning:
I have a situation in `brave-core` where I want to conditionally render some content in a dialog. I need to control this solely via CSS.

It works pretty well, except that the close event is fired on the dialog, even when it is `display: none`


AlanBreck (approved):
I don't fully understand the problem, but the solution looks good to me.

PR #1006, opened 2025-02-26
[Dialog]: Set overflow hidden - dialogs shouldn't scroll

fallaciousreasoning:


AlanBreck (commented):
I'm not sure we want this. For instance, in landscape mode on mobile, we want the dialog to scroll.

fallaciousreasoning:
I think the whole Dialog should scroll, but the content of the dialog shouldn't scroll

AlanBreck:
> I think the whole Dialog should scroll, but the content of the dialog shouldn't scroll

I agree, but this makes it so the dialog won't scroll, yes? Try viewing one of the stories in landscape mobile.

PR #1015, opened 2025-03-06
[Components]: Change instances of ':global selector' to ':global(selector)'

AlanBreck:
Using :global without parentheses around the targeted selector seems to
be undocumented, and results in an error in some environments. See for
example https://svelte.dev/playground/568dfa21589d4854a86243373526127c?version=4.2.19

### DIFF
--- a/src/components/icon/icon.svelte
+++ b/src/components/icon/icon.svelte
@@ -65,7 +65,7 @@
     flex-shrink: 0;
 
     & .icon,
-    :global svg {
+    :global(svg) {
       width: 100%;
       height: 100%;
     }
--- a/src/components/segmentedControl/segmentedControl.svelte
+++ b/src/components/segmentedControl/segmentedControl.svelte
@@ -169,35 +169,35 @@
         left 0.4s cubic-bezier(0.22, 1, 0.36, 1);
     }
 
-    :where(&) > :global .leo-control-item,
-    :where(&) > :global ::slotted(leo-controlitem) {
+    :where(&) > :global(.leo-control-item),
+    :where(&) > :global(::slotted(leo-controlitem)) {
       --leo-control-item-icon-color: var(--leo-color-icon-default);
       --leo-control-item-color: var(--leo-color-text-secondary);
       --leo-control-item-background: transparent;
       --leo-control-item-radius: calc(var(--radius) - var(--control-padding));
     }
 
-    :where(&:not(.transitioning)) > :global .leo-control-item:hover,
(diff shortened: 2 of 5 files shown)


src/components/icon/icon.svelte
```diff
@@ -65,7 +65,7 @@
     flex-shrink: 0;
 
     & .icon,
-    :global svg {
+    :global(svg) {
```
fallaciousreasoning, line 68: do we actually want this?

src/components/segmentedControl/segmentedControl.svelte
```diff
@@ -169,35 +169,35 @@
         left 0.4s cubic-bezier(0.22, 1, 0.36, 1);
     }
 
-    :where(&) > :global .leo-control-item,
-    :where(&) > :global ::slotted(leo-controlitem) {
+    :where(&) > :global(.leo-control-item),
```
fallaciousreasoning, line 172: Does this do anything when we're in a WebComponent?

AlanBreck: You mean does it break anything?

src/components/icon/icon.svelte
AlanBreck: I think yes. Why not?

src/components/segmentedControl/segmentedControl.svelte
fallaciousreasoning: I mean, does this component work as a webcomponent properly - I don't think these selectors will work

AlanBreck: The selector is working in the `plain-html` example (the greyed out variable names seems to be a dev tools quirk with the shadow dom).

SCR-20250310-kvxm-2

fallaciousreasoning: Sweet!

src/components/icon/icon.svelte
fallaciousreasoning: Doesn't it not need to be a child of this component though - that selector is for 
```css
svg {
  /* global svg styles */
```

So if I have an SVG which isn't in this component it'll still get these styles?

AlanBreck: No, because it's still nested within `.leoIcon`, so the output is:

```css
.leoIcon .icon,
.leoIcon svg {
  width: 100%;
  height: 100%;
}
```

fallaciousreasoning: :facepalm:

fallaciousreasoning (approved):

PR #1037, opened 2025-03-28
fix: autopin dependencies

thypon:
fix: autopin dependencies

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e # v4
+      - uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e # v4.3.0
         id: setup-node
         with:
           node-version: '22.14.0'
@@ -84,7 +84,7 @@ jobs:
         run: npm run build-storybook
 
       - name: Upload Artifact
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: build
           if-no-files-found: error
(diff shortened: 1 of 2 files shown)


.github/workflows/transform-tokens.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       # this creates a json file from the stringified json from the request body
       - name: create json from request body
         id: create-json
-        uses: jsdaniell/create-json@b8e77fa01397ca39cc4a6198cc29a3be5481afef # v1.2.3
+        uses: jsdaniell/create-json@b8e77fa01397ca39cc4a6198cc29a3be5481afef # v.1.2.3
```
fallaciousreasoning, line 25: nit - this should be reverted - just formats the version incorrectly and doesn't change the revision

.github/workflows/build.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         run: npm run build-storybook
 
       - name: Upload Artifact
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
```
fallaciousreasoning, line 87: interesting - the hash here hasn't changed but the version has. Is this intentional?

fallaciousreasoning: Is the `# v4` pinning to the major version of 4, while this will pin to the exact version?

fallaciousreasoning (commented):
this lgtm except for the change to `.github/workflows/transform-tokens.yml` which I think should be reverted as it looks like the version is formatted incorrectly.

I take it these comments are actually pinning the version, rather than just commenting which version is being used? If so that's kinda wild.

The audit failure is unrelated and is fixed by https://github.com/brave/leo/pull/1036?notification_referrer_id=NT_kwDOAHUoSLMxNTU1NDg4NDE1NDo3Njc4MDI0

If you rebase the audit should pass :smile:

.github/workflows/transform-tokens.yml
thypon: it is actually correct and incorrect at the same time. Looks like the author pushed a wrong tag

Screenshot 2025-03-27 at 19 11 48

.github/workflows/build.yml
thypon: yup. It will also allow for semver bumps inside `renovate`

.github/workflows/transform-tokens.yml
fallaciousreasoning: lol :laughing:

fallaciousreasoning: well thats terrifying

PR #1065, opened 2025-04-22
[types]: Fix generated WebComponent types

fallaciousreasoning:
`path.resolve` is relative to the package root, while `require.resolve` is relative to `node_modules`


petemill (approved):
Nicely figured out

AlanBreck:
Thanks, @fallaciousreasoning! Sorry for causing the bug.

Issue #997, opened 2025-02-21
[Dialog]: Incorrect tab order


AlanBreck:
https://github.com/user-attachments/assets/ed657797-984a-4dfc-a1f1-572ca15ff9b5

benjaminknox:
👆🏻Believe I have a fix for this in https://github.com/brave/leo/pull/1071, let me know your thoughts

PR #1071, opened 2025-04-30
[Dialog]: Fix input field tab order

benjaminknox:
# What?

Related tasks: https://github.com/brave/leo/issues/997

On the dialog, the tab order is incorrect when you have the close button turned on. In the case of the issue reported above (https://github.com/brave/leo/issues/997) the input field has the attribute `tabindex=1`,  and ~~adding the same `tabindex=1` to the close button keeps the order correct~~.

The bug is actually a side effect of adding `tabindex=1` to the Input field so tabbing works properly in a shadowroot for the input field. The fix here is to conditionally add tab index when the root node for the Input field is not the document.

## Details
- Conditionally adds `tabindex=1` attribute to the Input component when the root note is not the document
- Adds a storybook story for `Show Close`
- Adds an input field to the default Dialog component in storybook

### Before

https://github.com/user-attachments/assets/66d74a08-eca8-479a-b17c-d5a6be3965bd

### After

https://github.com/user-attachments/assets/3ce82aa4-5538-4a8c-a139-cbd41a032c60

### Input Field in the Shadow DOM
image

### DIFF
--- a/src/components/input/input.svelte
+++ b/src/components/input/input.svelte
@@ -138,7 +138,9 @@
   let tabindex : number;
   
   onMount(() => {
-    tabindex = input.getRootNode() === document ? $$restProps.tabindex : 1;
+    const inputRootNode = input.getRootNode() as ShadowRoot;
+
+    tabindex = inputRootNode.host?.tagName === 'LEO-INPUT' ? 1 : $$restProps.tabindex;
   });
 
 </script>
--- a/src/components/input/input.svelte
+++ b/src/components/input/input.svelte
@@ -14,6 +14,7 @@
   import Button from '../button/button.svelte'
   import FormItem, { type Mode, type Size } from '../formItem/formItem.svelte'
   import Icon from '../icon/icon.svelte'
+  import  { onMount } from 'svelte'
 
   type OverrideProps = 'type' | 'value' | 'size' | 'class' | `on:${string}`
   type LeoInputTypeAttribute =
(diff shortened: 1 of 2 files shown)


fallaciousreasoning (approved):
lgtm, thanks @benjaminknox :smile:

fallaciousreasoning:
Hey actually @benjaminknox I was having a chat with @AlanBreck and we reckon a better solution would be to only add the `tabindex` to `Input` if its inside a shadowRoot.

I think we could probably check in the `Input` on initial attach with:

`node.getRootNode() === document` (or something similar)

benjaminknox:
@fallaciousreasoning Added some updates, let me know your thoughts!

src/components/input/input.svelte
```diff
@@ -133,6 +134,13 @@
     value = e.currentTarget['value']
     hasErrorsInternal = (required && !value) || !input?.checkValidity()
   }
+
+  let tabindex : number;
+  
+  onMount(() => {
+    tabindex = input.getRootNode() === document ? $$restProps.tabindex : 1;
```
fallaciousreasoning, line 141: oh nice, I hadn't considered passing through `tabindex` from `$$restProps`! One suggestion, lets check its a `leo-input` in particular, in case we render this inside a different shadowRoot.

```suggestion
    tabindex = input.getRootNode().host?.tagName !== 'LEO-INPUT' ? $$restProps.tabindex : 1;
```

benjaminknox: Oh nice, yeah thanks for pointing that out. Done!

Went about it this way, think it reads a little better:

```ts
  onMount(() => {
    const inputRootNode = input.getRootNode() as ShadowRoot;

    tabindex = inputRootNode.host?.tagName === 'LEO-INPUT' ?  1 : $$restProps.tabindex;
  });

```

AlanBreck:
Thanks so much, @benjaminknox!

PR #1088, opened 2025-05-15
[Icons]: Fix update script

fallaciousreasoning:
I think this probably broke when we converted a bunch of stuff over to esmodules. I also noticed there was a timeout happening occasionally, so I increased the timeout to 10 minutes (there are some big API requests).

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "babel-loader": "9.2.1",
     "css-loader": "6.11.0",
     "css-tree": "2.3.1",
-    "figma-api-exporter": "github:brave/figma-api-exporter#08ea66f0794fcb3fa0240692d9016a562b4fbcef",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -62,7 +62,8 @@
     "svg2vectordrawable": "2.9.1",
     "svgo": "3.3.2",
     "ts-jest": "29.3.2",
-    "tsx": "4.19.4"
+    "tsx": "4.19.4",
+    "undici": "7.9.0"
```
fallaciousreasoning, line 66: this library is used internally by Node for HTTP requests, so we need it as a `devDependency` to modify the timeout

AlanBreck: Why did we need to modify the timeout?

AlanBreck (approved):

package.json
fallaciousreasoning: Possibly because I got rid of `axios` in favor of the built-in `fetch` which probably has a different default timeout

PR #1105, opened 2025-05-30
Token updates for colors and labels

aguscruiz:


src/tokens/universal.tokens.json
```diff
@@ -794,17 +794,17 @@
...
+          "fontWeight": 400,
           "fontStyle": "normal",
           "fontStretch": "normal",
           "letterSpacing": 0,
-          "lineHeight": 10,
+          "lineHeight": 12,
           "paragraphIndent": 0,
           "paragraphSpacing": 0,
-          "textCase": "uppercase"
+          "textCase": "none"
```
AlanBreck, line 807: @aguscruiz, the default value for text casing on the label component is `uppercase`, but doesn't come from the tokens. If this is changing, would you mind opening an issue to request that change in the component?

src/tokens/universal.variables.json
```diff
@@ -2783,14 +2787,6 @@
...
           "referencedVariable": "$schemes.primary"
         },
-        "hover-monochrome": {
-          "type": "color",
-          "value": "rgba(0, 0, 0, 0.1000)"
-        },
-        "hover-primary": {
-          "type": "color",
-          "value": "rgba(69, 81, 209, 0.1000)"
-        },
```
AlanBreck, line 2793: Is this not being used anywhere?

src/tokens/universal.tokens.json
AlanBreck: ...or make the change yourself in the label component on line 77.

src/tokens/universal.variables.json
aguscruiz: No, it's a color token I had added just for design and use in Figma mocks for buttons, but I solved it in a better way without needing these

src/components/label/label.svelte
```diff
@@ -74,7 +74,6 @@
     padding: var(--padding);
     font: var(--font-text);
     height: var(--leo-icon-m);
-    text-transform: var(--leo-label-capitalization, uppercase);
```
AlanBreck, line 77: @aguscruiz we should keep this, but switch `uppercase` to `none`.

src/tokens/universal.tokens.json
AlanBreck: Yep, those come from the tokens.

AlanBreck (approved):
LGTM!

aguscruiz:
Let's goooo, thanks @AlanBreck

Issue #705, opened 2024-06-06
Link should get a <button variant


petemill:
Link component should be allowed to render `<button` so that we can avoid developers using `<Link href="#"`. 

This plan is approved by @AlanBreck and @fallaciousreasoning

fallaciousreasoning:
This should involve rendering a 

```
<svelte:element this={isButton ? 'button' : 'a'}>
    ...
</svelte:element>
```

https://learn.svelte.dev/tutorial/svelte-element

jonathansampson:
Instead of `svelte:element`, would it not be easier to pull in https://github.com/brave/leo/blob/main/src/components/button/button.svelte to return an actual button when `isButton` is set/true? Otherwise, it seems we'll have quite a bit of work to maintain button-aesthetics across the two components.

fallaciousreasoning:
I think the point of this one is to not maintain the button aesthetics - it should always look like a link here

PR #1114, opened 2025-06-09
[Link]: Allow Link to be a button

AlanBreck:
Resolves https://github.com/brave/leo/issues/705

### DIFF
--- a/src/components/link/link.svelte
+++ b/src/components/link/link.svelte
@@ -1,32 +1,59 @@
 <script lang="ts">
   import type { SvelteHTMLElements } from 'svelte/elements'
 
-  type ExcludedProps = 'class' | 'href' | 'aria-disabled'
-  type $$Props = Omit<Partial<SvelteHTMLElements['a']>, ExcludedProps> & {
+  type Href = $$Generic<string | undefined>
+  type ExcludedProps = 'class' | 'aria-disabled' | 'href' | 'hreflang'
+
+  interface CommonNalaLinkProps {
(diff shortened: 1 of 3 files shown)


src/components/link/link.svelte
```diff
@@ -1,32 +1,59 @@
...
-  aria-disabled={isDisabled}
-  on:click={(e) => {
-    if (isDisabled) e.preventDefault()
-  }}
+  rel={href && 'noopener'}
+  href={href || undefined}
+  class="leoLink"
+  class:disabled
+  aria-disabled={disabled || undefined}
+  on:click={onClick ||
```
fallaciousreasoning, line 41: Should we still be firing onClick if we're disabled?

src/components/link/link.svelte
```diff
@@ -1,32 +1,59 @@
...
-  class="leo-link"
-  class:disabled={isDisabled}
-  aria-disabled={isDisabled}
-  on:click={(e) => {
-    if (isDisabled) e.preventDefault()
-  }}
+  rel={href && 'noopener'}
+  href={href || undefined}
+  class="leoLink"
+  class:disabled
```
fallaciousreasoning, line 39: I always get a bit confused - is this conditionally adding the class?

fallaciousreasoning (commented):
So the button version of the link can't have a link?

AlanBreck:
> So the button version of the link can't have a link?

Right

fallaciousreasoning (approved):

PR #1122, opened 2025-06-16
Token updates

aguscruiz:
This PR contains:

- Changes to font stack: Removes Inter and Poppins completely in favor of using SanFrancisco, which should be replaced in Nala for the default sans system font. 
- Some minor value updates for using system fonts
- Removed the "mobile" variable mode, and instead added a "search" one that's only for design, no code update needed.


fallaciousreasoning:
So just to confirm @aguscruiz these should be using `system-ui` instead of the system font right? We'll probably need to tweak the token generator a bit to handle that (as it looks like we're spitting out the MacOS system font)

aguscruiz:
Yeah, system-ui. We could do something like this to be sure we always default to the sans-serif default.
https://css-tricks.com/snippets/css/system-font-stack/

```
body {
  font-family: system-ui, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
}
```

PR #1143, opened 2025-07-03
responsive padding on dialog

aguscruiz:
Adjusted the padding on the dialog component to default to a smaller size when the screen size is smaller. This is handy in mobile and also browser panels


src/components/dialog/dialog.svelte
```diff
@@ -265,7 +265,7 @@
     gap: var(--leo-spacing-l);
   }
 
-  @media (min-width: 375px) and (min-height: 375px) {
+  @media (min-width: 480px) and (min-height: 375px) {
```
fallaciousreasoning, line 268: I think this is to work in landscape/portrait on mobile
```suggestion
  @media (min-width: 480px) and (min-height: 480px) {
```

fallaciousreasoning (approved):

PR #1176, opened 2025-08-21
[Dropdown]: ensure inline dropdown is actually inline

AlanBreck:


src/components/formItem/formItem.svelte
```diff
@@ -297,6 +298,10 @@
     display: flex;
     flex-direction: row;
     gap: var(--leo-spacing-s);
+
+    &:empty {
```
fallaciousreasoning, line 302: did you test this works okay in the webcomponent?

fallaciousreasoning (approved):

src/components/formItem/formItem.svelte
AlanBreck: No, but I will. Pretty sure we've tested this pattern in other cases, though.

AlanBreck: Confirmed.

PR #1177, opened 2025-08-22
[FormItem]: fix sizing/v-alignment for inline mode

AlanBreck:


src/components/dropdown/dropdown.stories.svelte
```diff
@@ -142,6 +142,7 @@
   <Link>Footer link</Link>
   <Link>Footer link</Link>
   <Dropdown value={undefined} {...args} mode="inline">
+    <Icon name="country-us" slot="left-icon" />
```
fallaciousreasoning, line 145: huh interesting - I thought it wasn't kosher to slot a svelte element into a slot (and that you had to wrap it in a div).

Maybe they fixed it in Svelte 4?

src/components/formItem/formItem.svelte
```diff
@@ -255,6 +255,8 @@
     --border-color: transparent;
     --border-color-hover: transparent;
     --shadow-hover: none;
+    --font: inherit;
+    --leo-icon-size: 1.43em;
```
fallaciousreasoning, line 259: maybe a comment explaining where this magic number comes from (`20px default icon size / 14px default font-size`)

fallaciousreasoning (approved):

src/components/dropdown/dropdown.stories.svelte
AlanBreck: Maybe? I wasn't aware of it not being preferred.

fallaciousreasoning: I think it was a compile error in Svelte 3

src/components/formItem/formItem.svelte
AlanBreck:
```suggestion
    /**
     * This number is a ratio to preserve the proportions between
     * icon and text found the `normal` size:
     * 20px 'normal; icon size / 14px 'normal' font-size
     */
    --leo-icon-size: 1.43em;
```

PR #1179, opened 2025-08-26
[FormItem]: Fix layout bug introduced by '[FormItem: Fix spacing'

AlanBreck:
| Before | After |
|--------|--------|
| Screenshot 2025-08-25 at 10 13 52 PM | Screenshot 2025-08-25 at 10 16 01 PM | 

Introduced by https://github.com/brave/leo/commit/4924dde3bf6d1f9f458a7335a3e883b349e4e635

### DIFF
--- a/src/components/formItem/formItem.svelte
+++ b/src/components/formItem/formItem.svelte
@@ -165,7 +165,6 @@
 
     .content {
       flex: 1;
-      width: 100%;
       padding: 0 var(--leo-spacing-s);
     }
 
--- a/src/components/formItem/formItem.svelte
+++ b/src/components/formItem/formItem.svelte
@@ -93,9 +93,7 @@
 >
   {#if $$slots.label && renderLabel}
     <div class="label-row">
-      <slot name="label" />{#if required}<span class="required-indicator"
-          >*</span
-        >{/if}
+      <slot name="label" />{#if required}<span class="required-indicator">*</span>{/if}
     </div>
   {/if}
   <div class="control" bind:this={controlElement}>
--- a/src/components/formItem/formItem.svelte
+++ b/src/components/formItem/formItem.svelte
@@ -165,13 +165,8 @@
 
     .content {
       flex: 1;
-
-      display: flex;
-      flex-direction: column;
-      justify-content: stretch;
+      width: 100%;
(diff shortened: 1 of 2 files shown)


src/components/formItem/formItem.svelte
```diff
@@ -93,7 +93,9 @@
 >
   {#if $$slots.label && renderLabel}
     <div class="label-row">
-      <slot name="label" />{#if required}<span class="required-indicator">*</span>{/if}
+      <slot name="label" />{#if required}<span class="required-indicator"
+          >*</span
```
fallaciousreasoning, line 97: does it still pass `npm run format` if you revert this? Its ugly as 😆

src/components/formItem/formItem.svelte
```diff
@@ -161,12 +163,17 @@
...
+      flex: 1;
+
+      display: flex;
+      flex-direction: column;
+      justify-content: stretch;
       padding: 0 var(--leo-spacing-s);
-      margin-inline-end: auto;
+
+      cursor: text;
     }
```
fallaciousreasoning, line 177: can we not just set `width: 100%` on the children of this element, rather than making it a flex?

AlanBreck: We probably could. I basically just copied what we had over in input.

fallaciousreasoning:
cc @szilardszaloki who just encountered this in the wild

src/components/formItem/formItem.svelte
```diff
@@ -161,12 +161,12 @@
     justify-content: stretch;
     font: var(--font);
     gap: var(--gap);
-    transition:
-      color 0.2s ease-in-out;
+    transition: color 0.2s ease-in-out;
 
     .content {
+      flex: 1;
+      width: 100%;
```
fallaciousreasoning, line 168: hey, just tested and I don't think we need this `width: 100%` - it comes from the flex: 1

fallaciousreasoning (commented):
one more nit and lgtm

fallaciousreasoning (approved):

src/components/formItem/formItem.svelte
fallaciousreasoning:
```suggestion
```

PR #1181, opened 2025-08-28
[Checkbox]: Make checkbox WebComponent work on iOS

fallaciousreasoning:
This seems to be because WebKit doesn't like having a `:has` selector in a `shadowRoot`, which is why the Storybook was working


Brandon-T (approved):

petemill (approved):
Works well in safari. Did you find a webkit bug or some other tracker for this?

PR #1213, opened 2025-09-16
[ButtonMenu]: Fix margin for single item

fallaciousreasoning:
When there was just a single item the bottom margin was overriding the top one. Solution is to set the top/bottom padding separately.


aguscruiz (approved):
Looks good!

AlanBreck (approved):

PR #1215, opened 2025-09-17
[Toggle]: Don't use Sass

fallaciousreasoning:
The recent bump to Sass broke our `theme` plugin. Fortunately, modern CSS has most of the features we'd use from Sass so converting over to plain CSS fixes things!

### DIFF
--- a/src/components/toggle/toggle.svelte
+++ b/src/components/toggle/toggle.svelte
@@ -91,7 +91,7 @@
   <slot {checked} />
 </label>
 
-<style lang="scss">
+<style>
   :global(:root) {
     --leo-direction: 1;
   }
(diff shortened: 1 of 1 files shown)


src/components/toggle/toggle.svelte
```diff
@@ -150,6 +143,18 @@
     -webkit-tap-highlight-color: transparent;
   }
 
+  @media (prefers-reduced-motion) {
```
fallaciousreasoning, line 146: note: modern CSS doesn't support nesting media queries so we have to unnest them

bsclifton:
For example of how toggle broken, see https://github.com/brave/brave-browser/issues/49379

AlanBreck:
I'll check our website analytics to ensure we can support CSS nesting.

PR #1223, opened 2025-09-26
[Tailwind]: Prevent postcss-js from merging at-rules in extracted component styles

AlanBreck:
Since objects can't have duplicate keys, the process of converting our CSS rules to JS objects resulted in `postcss-js` combining multiple `at-rules` into one, which resulted in improper ordering of queries and styles.


fallaciousreasoning:
Man, I'm quite surprised that sass change wasn't listed as breaking :D

Issue #1239, opened 2025-10-15
Button Menu does not respect --leo-menu-control-width variable


Douglashdaniel:
If you attempt to use the `--leo-menu-control-width` css variable to set the `Width` of a `Button Menu`, nothing currently happens.

AlanBreck:
Mentioned this in the PR, but the intent of that variable is to be read only.

PR #1240, opened 2025-10-15
[Button Menu]: Fix --leo-menu-control-width Variable

Douglashdaniel:
Fixes a bug where the `Button Menu` would not respect the `--leo-menu-control-width` variable

Resolves <https://github.com/brave/leo/issues/1239>

### DIFF
--- a/src/components/menu/menu.svelte
+++ b/src/components/menu/menu.svelte
@@ -45,6 +45,7 @@
   export let onClose: CloseEvent =
     undefined
   export let onSelectItem: (detail: SelectItemEventDetail) => void = undefined
+  export let fullWidth: boolean = true
 
   function dispatchClose(
     originalEvent: Event,
@@ -97,8 +98,11 @@
 
   // Note: we check isOpen !== undefined here so this is recalculated every time the
   // dropdown is opened/closed.
-  $: minWidth =
-    (isOpen !== undefined && target?.getBoundingClientRect().width) || 0
+  $: minWidth = fullWidth 
+    ? (isOpen !== undefined && target?.getBoundingClientRect().width) || 0
+    : 0
+
+
 
   function selectMenuItem(e: Event) {
     // Find the option which was clicked on, if any.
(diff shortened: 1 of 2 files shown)


src/components/menu/menu.svelte
```diff
@@ -176,6 +180,7 @@
 
   function applySizeMiddleware({ rects, availableHeight }) {
     popup.style.maxHeight = `var(--leo-menu-max-height, calc(${availableHeight}px - var(--leo-spacing-xl)))`
+    popup.style.minWidth = `var(--leo-menu-control-width, ${minWidth}px)`
```
fallaciousreasoning: is there a functional difference, doing this here, rather than setting the style directly?

fallaciousreasoning: Doesn't it get applied to the same element?

fallaciousreasoning (commented):
Hey just want to check I'm understanding this right:

`--leo-menu-control-width` is being ignored when you set it on a `ButtonMenu` because `ButtonMenu` sets the variable on itself, right?

I have a vague recollection of discussing this with @alanbreck at the time and feeling like it was weird - I think he wanted it as a CSS variable that children could use to tell how wide the control was.

Is there a way to get this working without the `fullWidth` property? What about if we split the CSS property into two?

`--leo-menu-control-width` for controlling the min-width
`--leo-menu-control-natural-width` for letting children know the size (via the weird calculation)

and then 

`min-width: var(--leo-menu-control-width, var(--leo-menu-control-natural-width))`

so we have the most expected behavior.

Then we should label this PR with a **BREAKING** at the start of the commit (I don't think much actually uses this).

AlanBreck:
This was intended as a read only property. I thought I had a comment in there, but I'm AFK right now and can't check. I believe there's a note about it in Storybook.

Douglashdaniel:
Happy go whatever direction you guys suggest here.
I'm trying to fix the `Tools Menu` in `Leo AI` from always being full width.. maybe I've overlooked something in `brave-core`. @fallaciousreasoning you may have some insight haha

Screenshot 20

PR #1256, opened 2025-11-15
[github workflows] add permissions

thypon:


.github/workflows/build.yml
```diff
@@ -9,6 +9,9 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
 
+permissions:
+  contents: read
```
fallaciousreasoning: This action also:
1. Writes comments on a PR
2. Uploads files to AWS

.github/workflows/test-tokens.yml
fallaciousreasoning: @AlanBreck does this workflow ever run?

.github/workflows/sf-symbols.yml
```diff
@@ -8,6 +8,9 @@ on:
       - main
   workflow_dispatch:
 
+permissions:
+  contents: read
```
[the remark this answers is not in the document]
thypon: network is always granted, moreover to trigger the other workflow, it's using a github token in the `secrets` . This permission only controls the default `GITHUB_TOKEN` scope.

https://github.com/brave/devops/wiki/GitHub-Actions-Permissions-Reference

.github/workflows/update-icons.yml
```diff
@@ -8,6 +8,9 @@ on:
 env:
   FIGMA_API_TOKEN: ${{ secrets.FIGMA_API_TOKEN }}
 
+permissions:
+  contents: read
```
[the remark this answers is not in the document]
thypon: Needs `pull-requests: write` indeed

.github/workflows/test-tokens.yml
AlanBreck: Not that I'm aware of. Appears to have been boilerplate from when we initially configured the design tokens Figma plugin.

.github/workflows/build.yml
fallaciousreasoning: Uploading the files is line 94 in this file. Commenting actually seems to be in a separate workflow `deploy` which probably needs permissions added too

fallaciousreasoning:
Looks like we're missing permissions on deploy.yaml still

.github/workflows/update-icons.yml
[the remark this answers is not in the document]
thypon: fixex in ee8a9297

PR #1300, opened 2026-02-02
[Dropdown]: Improve default value handling

fallaciousreasoning:
This makes it much easier to use the dropdown
component as a drop in replacement for a select.

Additionally it means the dropdown just "Does the
right thing" most of the time.

### DIFF
--- a/examples/plain-html/index.html
+++ b/examples/plain-html/index.html
@@ -232,7 +232,6 @@
     <leo-button id="dialog-button">Show Dialog</leo-button>
     <leo-dropdown placeholder="Select an option..." value="NZ">
       <leo-icon slot="left-icon" name="flag"></leo-icon>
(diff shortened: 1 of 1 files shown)


examples/plain-html/index.html
```diff
@@ -230,9 +230,9 @@
       </div>
     </leo-alert>
     <leo-button id="dialog-button">Show Dialog</leo-button>
-    <leo-dropdown placeholder="Select an option...">
+    <leo-dropdown placeholder="Select an option..." value="NZ">
       <leo-icon slot="left-icon" name="flag"></leo-icon>
-      <div id="selectedCountry" slot="value"></div>
+      <div id="selectedCountry"></div>
```
AlanBreck: Is this necessary at all?

Issue #1306, opened 2026-02-10
Can't call .click() on leo-dropdown


Douglashdaniel:
Currently are not able to call .click() on "leo-dropdown"

PR #1307, opened 2026-02-10
Add .click() on leo-dropdown

Douglashdaniel:
Resolves <https://github.com/brave/leo/issues/1306>


fallaciousreasoning:
Ah, the reason its click here is so it gets called when you do `document.querySelector('leo-dropdown').click()

PR #1319, opened 2026-02-25
[Alert]: Add 'small' and 'inline actions' variants

AlanBreck:
Related brave-core issue:
https://github.com/brave/brave-browser/issues/52891

NOTE: I'm not married to the name "thin" to describe the one variant. It's marked in Figma as "side actions", but that's a bit limiting. I'd be inclined to call it "inline", but that name is taken in Figma by the opposite of `toast`.


fallaciousreasoning (approved):
lgtm % nits

PR #1350, opened 2026-04-01
Improve CI for sec dependency bumps

kdenhartog:
- This adds a socket fix CI to make it easier to bump dependencies.
- Additionally it makes it so the CI runs with the audit-config list like what we do in brave-core for consistency.


kdenhartog:
This CI didn't work when I ran it previously on WDP. Gonna fix it before merging this

PR #1352, opened 2026-04-01
chore: enforce strict node and npm engine versions

kdenhartog:
## Summary

- Adds `engines` field to `package.json` pinning Node to `>=24.11.1 <25.0.0` and npm to `>=11.6.2`
- Sets explicit `node-version: '24.11.1'` in the Socket Fix workflow
- Regenerates `package-lock.json` to reflect the engine constraints

## Why

Loose engine ranges allow CI and local environments to drift, which can cause subtle behavior differences between contributors and automated workflows. Pinning to exact minor versions ensures consistency.

## Test plan

- [ ] `npm install` succeeds with the pinned Node/npm versions
- [ ] CI passes on the updated `package-lock.json`


mihaiplesa (changes requested):

AlanBreck:
@kdenhartog, FYI we're hoping to move towards `pnpm`: https://github.com/brave/leo/pull/1394

kdenhartog:
Let's just do this when we're adding NPM then cause that will likely force this to be fixed throughout Leo/WDP/Brave-core at the same time too

PR #1373, opened 2026-04-21
[Menu]: Fix bug where menu wouldn't flip

fallaciousreasoning:


AlanBreck (commented):
The docs for floating UI seem to address our very issue here: https://floating-ui.com/docs/size#maxheight-style-left-on-floating-element

They're fix seems to work, when trying it locally:

```js
function applySizeMiddleware({ availableHeight }) {
  if (!popup?.style) return;
  popup.style.maxHeight =
    availableHeight >= popup.scrollHeight
      ? ''
      : `var(--leo-menu-max-height, calc(${availableHeight}px - var(--leo-spacing-xl)))`
}
```

This would avoid the need to set a min height, unless we still think that could be valuable.

AlanBreck (approved):

fallaciousreasoning:
Discussed offline about both approaches and going with this for now

PR #1379, opened 2026-05-01
[Link]: Add support for icons

AlanBreck:


src/components/link/link.stories.svelte
```diff
@@ -73,14 +74,52 @@
...
+    description="The link supports icon-before and icon-after slots in addition to the default text slot"
   >
     <Slot name="default" explanation="The text of the link">
       <div>
         <Link {...args} text={null}>{args.text}</Link>
       </div>
     </Slot>
+    <Slot name="icon-before" explanation="An icon displayed before the link text">
+      <div>
+        <Link {...args} text={null}>
```
fallaciousreasoning: why text=null here?

src/components/link/link.stories.svelte
```diff
@@ -73,14 +74,52 @@
...
+          <Icon name="check-circle-outline" slot="icon-before" />
+          {args.text}
+        </Link>
+      </div>
+    </Slot>
+    <Slot name="icon-after" explanation="An icon displayed after the link text">
+      <div>
+        <Link {...args} text={null}>
```
fallaciousreasoning: ditto

fallaciousreasoning (approved):
lgtm % bonus `text=null`s

src/components/link/link.stories.svelte
AlanBreck: This slipped through from Claude without me reviewing it. I'm embarrassed. However, it's replicating what's in an existing story here:

https://github.com/brave/leo/blob/6c69ad4f1082ab5e98736e183c4f6787ff068c6a/src/components/link/link.stories.svelte#L80

fallaciousreasoning: Yeah, don't think the existing thing makes sense either :D

AlanBreck: I'll remove all of them.

fallaciousreasoning (approved):

src/components/link/link.stories.svelte
AlanBreck: So, these _are_ actually needed because spreading `{..args}` (which includes the `text` property) sets the [HTMLAnchorElement.text property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/text).

src/components/link/link.stories.svelte
```diff
@@ -73,11 +74,27 @@
 
 <Story name="Slots" let:args>
   <SlotInfo
-    description="The link has a single slot, which works the same as the slot on the anchor tag"
+    description="The link supports icon-before and icon-after slots in addition to the default text slot"
   >
     <Slot name="default" explanation="The text of the link">
       <div>
-        <Link {...args} text={null}>{args.text}</Link>
+        Lorem ipsum dolor sit amet consectetur adipisicing elit <Link {...args} text={null}>{args.text}</Link>. Voluptatum magni nobis vel architecto tenetur repellendus dicta provident cumque doloribus illum voluptates, accusantium reprehenderit numquam voluptas ad autem neque? Cumque, consectetur.
```
fallaciousreasoning, line 81: lol text null

fallaciousreasoning (approved):
lgtm % text null

src/components/link/link.stories.svelte
fallaciousreasoning: lol :D

PR #1380, opened 2026-05-01
[Link]: Add support loading state

AlanBreck:


src/components/link/link.svelte
```diff
@@ -36,14 +39,26 @@
...
+      {#if $$slots.loading}
+        <slot name="loading" />
+      {:else}
+        <slot />
+      {/if}
+    </div>
+    <ProgressRing />
+  {:else}
+    <slot />
+  {/if}
```
fallaciousreasoning, line 61: do we need this wrapper div? It also seems like we can't replace the loading spinner with the loading slot?

should this be:


```suggestion
  {#if isLoading
      <slot name="loading">
        <ProgressRing />
      </slot>
    {/if}
    <slot />
```

fallaciousreasoning: ah okay, I checked and what you have here is pretty much exactly what we have in the button., so this is fine

fallaciousreasoning (approved):

PR #1381, opened 2026-05-05
[Link]: Fix loading state

AlanBreck:


src/components/link/link.svelte
```diff
@@ -39,22 +40,20 @@
...
   on:click={onClick ||
     ((e) => {
-      if (disabled) e.preventDefault()
+      if (loading || disabled) e.preventDefault()
     })}
-  disabled={isLoading || disabled || undefined}
+  disabled={disabled || undefined}
 >
   {#if isLoading}
-    <div class="content">
```
AlanBreck, line 51: This was superfluous.

src/components/link/link.svelte
```diff
@@ -81,6 +80,8 @@
     --focus-shadow: var(--leo-link-focus-shadow, var(--leo-effect-focus-state));
     --leo-icon-color: var(--leo-link-icon-color, currentColor);
     --leo-icon-size: var(--leo-link-icon-size, 1.3em);
+    --leo-progressring-size: var(--leo-icon-size);
+    --leo-progressring-color: var(--leo-icon-color);
```
AlanBreck, line 84: The rings are currently huge; this resolves that.

src/components/link/link.svelte
```diff
@@ -30,6 +30,7 @@
 
   $: tag = href ? 'a' : ('button' as 'a' | 'button')
   $: disabled = !!(isDisabled || (isDisabled as any) === '')
+  $: loading = !!(isLoading || (isLoading as any) === '')
```
AlanBreck, line 33: This is to keep consistent with the pattern used in `disabled`.

src/components/link/link.svelte
```diff
@@ -39,22 +40,20 @@
...
-  aria-disabled={disabled || undefined}
+  class:loading
+  aria-disabled={loading || disabled || undefined}
   on:click={onClick ||
     ((e) => {
-      if (disabled) e.preventDefault()
+      if (loading || disabled) e.preventDefault()
     })}
-  disabled={isLoading || disabled || undefined}
+  disabled={disabled || undefined}
```
AlanBreck, line 49: Loading state isn't necessarily the same as disabled, so this differentiates.

fallaciousreasoning: I think in the `Button` the loading selector just has more specificity than the disabled state, so its fine to also set disabled. Not sure if that applies here too.

fallaciousreasoning (approved):

src/components/link/link.svelte
AlanBreck: You mean from a CSS perspective?

AlanBreck: @fallaciousreasoning, if so, this proposed change isn't about the CSS rules, but rather that `disabled` isn't recommended for buttons in a loading state. Might not be a big deal, though.

fallaciousreasoning: makes sense :+1:

PR #1386, opened 2026-05-11
Navigation style updates

aguscruiz:
## Summary

This PR tightens navigation layout spacing and aligns nav item sizing with Leo’s button sizing.

## Changes

- **navigationMenu**: Removed vertical/horizontal spacing between items by dropping `gap: var(--leo-spacing-m)` on the `<ul>`. Items now sit flush with no gutter between adjacent `li` rows/columns.

- **navigationItem**: Tweaked hit area and inset for each link/button row:
  - Replaced fixed `height: 48px` with `min-height: 44px` so rows can grow with content while matching the 44px touch target pattern used elsewhere (e.g. medium buttons).
  - Reduced leading padding from `var(--leo-spacing-2xl)` to `var(--leo-spacing-xl)` so labels align closer to the left edge.

## How to verify

- Open Storybook (or consuming app): navigation vertical and horizontal entries should stack with **no gap** between items; each row should be at least **44px** tall with **xl** horizontal padding inside the clickable area.

### DIFF
--- a/src/components/navigation/navigationItem.svelte
+++ b/src/components/navigation/navigationItem.svelte
@@ -151,7 +151,7 @@
       box-sizing: border-box;
       gap: var(--leo-spacing-xl);
       align-items: center;
-      height: 44px;
+      min-height: 44px;
       padding-left: var(--leo-spacing-xl);
       padding-right: var(--leo-spacing-m);
       border-radius: 0;
--- a/src/components/navigation/navigationItem.svelte
+++ b/src/components/navigation/navigationItem.svelte
@@ -151,8 +151,8 @@
       box-sizing: border-box;
       gap: var(--leo-spacing-xl);
       align-items: center;
-      height: 48px;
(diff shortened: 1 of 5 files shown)


src/components/navigation/navigationItem.svelte
```diff
@@ -151,8 +151,8 @@
       box-sizing: border-box;
       gap: var(--leo-spacing-xl);
       align-items: center;
-      height: 48px;
-      padding-left: var(--leo-spacing-2xl);
+      min-height: 44px;
```
AlanBreck, line 154: Curious on the motivation for this. Will this lead to inconsistencies?

AlanBreck (approved):

src/components/navigation/navigationItem.svelte
aguscruiz: I just realized that the implementation didn't fully match Figma and I wasn't sure when it diverted, so let's see :D

PR #1391, opened 2026-05-13
chore(deps): updates deps and adds .npmrc with min-release-age

AlanBreck:


fallaciousreasoning (approved):

petemill (approved):
Confirmed with @AlanBreck that the packages were updated _after_ applying the `min-release-age` config

PR #1394, opened 2026-05-14
Migrate to `pnpm`

AlanBreck:
This also refactors away from `fs-extra` and `change-case` deps.


fallaciousreasoning:
I've tested and this seems to work fine in brave-core

package.json
```diff
@@ -33,12 +33,13 @@
...
     "@types/jsdom": "28.0.1",
+    "@types/node": "25.9.1",
     "@types/react": "18.3.28",
     "css-tree": "2.3.1",
     "figma-api-exporter": "github:brave/figma-api-exporter#141c3eab4643ba9cdcbb23867785aa7df9347eec",
+    "jest": "30.4.2",
     "jsdom": "28.1.0",
-    "lodash.camelcase": "4.3.0",
-    "lodash.merge": "4.6.2",
+    "lodash": "4.18.1",
```
fallaciousreasoning, line 42: do we want to pull in all of lodash? do we need it?

src/tokens/transformation/common/copyFileOrFolder.js
```diff
@@ -1,16 +1,16 @@
-import fs from 'fs-extra'
```
fallaciousreasoning, line 1: are we still pulling this in somewhere? Its not removed from package.json

src/tokens/utils.js
```diff
@@ -1,4 +1,5 @@
-import merge from 'lodash.merge'
+import words from 'lodash/words'
+import merge from 'lodash/merge'
```
fallaciousreasoning, line 2: could we remove our dependency on lodash? Maybe in a prereq PR. I feel like we could implement this ourselves and have one less dep

examples/example-ui-react/pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+allowBuilds:
+  esbuild: false
```
fallaciousreasoning, line 2: why do we need this?

src/tokens/transformation/ios/fontStyles.js
```diff
@@ -1,5 +1,5 @@
-import fs from 'fs-extra'
-import camelCase from '../common/camelCaseHelper'
+import fs from 'node:fs'
```
fallaciousreasoning, line 1: possibly removing `fs-extra` should happen in a prereq PR?

src/tokens/transformation/common/copyFileOrFolder.js
AlanBreck: It was never in `package.json`. It was getting hoisted by `npm` from some other dep, so I just migrated away.

src/tokens/utils.js
AlanBreck: We're relying on `lodash/template` in several places, so figured we might as well just use it for these too since the package is getting installed anyhow. This import style, though, allows for treeshaking.

package.json
AlanBreck: We use it for `lodash/template`. Lodash docs discourage use of "per method packages": https://lodash.com/per-method-packages.

examples/example-ui-react/pnpm-workspace.yaml
AlanBreck: `pnpm@11` requires explicit allow/deny lists for dependencies that want to run scripts.

https://pnpm.io/settings#allowbuilds

src/tokens/transformation/ios/fontStyles.js
AlanBreck: No strong feelings myself. Happy to pull out, if necessary.

fallaciousreasoning (approved):

PR #1402, opened 2026-05-29
[react]: Use useLayoutEffect instead of useEffect in svelte-react.ts

fallaciousreasoning:
This avoids a potential flash of the component with the default props before useEffect runs.


fallaciousreasoning:
cc @AlanBreck

AlanBreck (approved):

PR #1415, opened 2026-06-24

brave-support-admin:
> tsx ./src/scripts/update-icons.ts


AlanBreck (approved):

icons/social-bluesky.svg
AlanBreck, line 1: @aguscruiz, Not sure how much it matters, but any idea on why the background color is differe?

AlanBreck (approved):

icons/social-bluesky.svg
AlanBreck: Seems to be on account of the icon name (rightly) starting with `social-`:

https://github.com/brave/leo/blob/5eb91eb4bb9d62f55b062cc894f341b3502ac7ad/src/scripts/update-icons.ts#L57-L63

Don't think we need to block on this.

PR #1423, opened 2026-07-15
Add scrollable dialog & fix CSS selectors

aguscruiz:
## Summary

Fixes dialog actions layout/CSS bugs and improves scrollable dialog behavior (sticky chrome, scroll dividers, close button).

### Bug fixes
- **Actions missing background & padding** — Actions layout now handles slotted action content without opting `.leo-dialog` / `.actions` out of Svelte component scoping. Only the projected `[slot=actions]` content is wrapped in `:global(...)`.
- **Close button scrolling away** — With inner scroll, the absolutely positioned close button scrolled with the content. It now lives inside the sticky header when a header is present; without a header it uses sticky positioning so it stays pinned.

### Scrollable dialog behavior
- **Sticky actions** — Actions stay pinned to the bottom of the scrollport while the body scrolls.
- **Scroll dividers** — Header already showed a bottom border via `animation-timeline: scroll()` when scrolled. Actions now get a matching top border that appears when more content can scroll and fades at the end of the scroll range.
- **Actions gap** — Spacing between action children tightened from `--leo-spacing-xl` to `--leo-spacing-m`.

### Stories
- Added **Inner Scroll** story: height-capped dialog (`max-height` + `overflow: auto`), long body content, sticky header/actions with scroll dividers, close button, and Cancel/Done actions.
- Default template: moved the sample `Input` from the actions slot into the body (actions are buttons only).

## Test plan
- [ ] Open **Components / Dialog / Default** — actions sit inside the card with padding and background
- [ ] Open **Inner Scroll** — dialog scrolls within max-height; header and actions stay pinned
- [ ] On Inner Scroll, scroll the body — header bottom divider and actions top divider appear/update with scroll
- [ ] On Inner Scroll (and Show Close), close button stays fixed in the header while content scrolls
- [ ] Dialog without title/subtitle but with `showClose` — close still stays pinned while scrolling
- [ ] Short dialogs (Default / Show Close) — no stray actions top border when content does not overflow
- [ ] React / web-component consumers still get actions styling via `:host` rules
- [ ] Mobile vs desktop: actions stack in portrait / smaller sizes and row-align on larger viewports
- [ ] `npx jest`

### DIFF
--- a/src/components/dialog/dialog.svelte
+++ b/src/components/dialog/dialog.svelte
@@ -169,10 +169,7 @@
   /** Since Svelte 4 doesn't support conditional slots in the consumer,
    * we only want to account for actions if there's actually content in the slot
    * however for webcomponents, we don't need to check for this, so we special
-   * case the selector with :host.
-   *
-   * Only the slotted `[slot=actions]` content is `:global` (it's projected in
-   * by the parent). `.leo-dialog`/`.actions` stay component-scoped by Svelte. */
+   * case the selector with :host. */
   :host .leo-dialog.hasActions,
   .leo-dialog.hasActions:has(:global([slot=actions]:not(:empty))) {
     grid-template-rows: auto auto;
--- a/src/components/dialog/dialog.svelte
+++ b/src/components/dialog/dialog.svelte
@@ -59,15 +59,15 @@
       if (escapeCloses) close()
     }}
   >
-    {#if showClose}
-      <div class="close-button">
-        <Button kind="plain-faint" fab onClick={close}>
-          <Icon name="close" />
-        </Button>
-      </div>
-    {/if}
     {#if hasHeader}
       <header>
+        {#if showClose}
+          <div class="close-button">
+            <Button kind="plain-faint" fab onClick={close}>
+              <Icon name="close" />
+            </Button>
+          </div>
+        {/if}
         {#if showBack || $$slots.title}
           <div class="title-row">
             {#if showBack}
@@ -88,6 +88,12 @@
           </div>
         {/if}
       </header>
+    {:else if showClose}
+      <div class="close-button">
+        <Button kind="plain-faint" fab onClick={close}>
+          <Icon name="close" />
+        </Button>
+      </div>
     {/if}
     <div class="body">
       <slot />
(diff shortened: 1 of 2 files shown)


src/components/dialog/dialog.svelte
```diff
@@ -88,6 +88,12 @@
           </div>
         {/if}
       </header>
+    {:else if showClose}
+      <div class="close-button">
+        <Button kind="plain-faint" fab onClick={close}>
+          <Icon name="close" />
+        </Button>
+      </div>
```
fallaciousreasoning, line 96: @AlanBreck this is why I like React 😆 - is there a way to assign some fragment to a variable?

src/components/dialog/dialog.svelte
```diff
@@ -245,17 +265,36 @@
...
+  :global(.leo-dialog .actions:has([slot=actions]:not(:empty))) {
     background: var(--background);
     padding: var(--padding);
+    position: sticky;
+    bottom: 0;
+    z-index: 1;
+
+    border-top: 1px solid transparent;
+
+    @supports (animation-timeline: scroll()) {
```
fallaciousreasoning, line 281: ooooh fancy!

fallaciousreasoning (approved):

src/components/dialog/dialog.svelte
AlanBreck: I think you can do that with snippets in runes mode, right? Not precisely sure how this affects consumers, though.

src/components/dialog/dialog.svelte
```diff
@@ -163,14 +169,18 @@
...
    * however for webcomponents, we don't need to check for this, so we special
-   * case the selector with :host. */
+   * case the selector with :host.
+   *
+   * Use `:global(.foo:has(...))` (not `:has(:global(...))`) so `:global` is
+   * stripped correctly — nested `:global` inside `:has` can leak into the CSS
+   * as an invalid rule and drop background/padding on `.actions`. */
   :host .leo-dialog.hasActions,
-  .leo-dialog.hasActions:has(:global([slot=actions]:not(:empty))) {
+  :global(.leo-dialog.hasActions:has([slot=actions]:not(:empty))) {
```
AlanBreck: I don't think we actually want this entire selector to be global. Not understanding why this is necessary.

src/components/dialog/dialog.svelte
```diff
@@ -163,14 +169,18 @@
...
     grid-template-rows: auto auto;
   }
 
   :host .leo-dialog.hasHeader.hasActions,
-  .leo-dialog.hasHeader.hasActions:has(.actions :global([slot=actions]:not(:empty))) {
+  :global(.leo-dialog.hasHeader.hasActions:has(.actions [slot=actions]:not(:empty))) {
```
AlanBreck: Same here.

src/components/dialog/dialog.svelte
```diff
@@ -245,17 +265,36 @@
...
-  // selector follows the `:has()`. jsdom's nwsapi selector engine fails to parse selectors
-  // of the form `<X>:has(...) <Y>:where(...)` that Svelte 5 generates for scoped CSS.
-  .leo-dialog.hasActions .body:has(~ .actions :global([slot=actions]:not(:empty))) {
+  :global(.leo-dialog.hasActions .body:has(~ .actions [slot=actions]:not(:empty))) {
     padding-bottom: 0;
   }
 
   :host .leo-dialog .actions,
-  .leo-dialog .actions:has(:global([slot=actions]:not(:empty))) {
+  :global(.leo-dialog .actions:has([slot=actions]:not(:empty))) {
```
AlanBreck: Same here.

src/components/dialog/dialog.svelte
aguscruiz: Good catch — reverted. Only the slotted `[slot=actions]` content needs to be `:global` (it's projected in by the parent); `.leo-dialog` / `.actions` stay component-scoped again. Confirmed the nested `:has(:global(...))` form still compiles to valid CSS on our Svelte version.

src/components/dialog/dialog.svelte
aguscruiz: Same fix applied here — only the slotted content is `:global` now.

src/components/dialog/dialog.svelte
aguscruiz: Same fix applied here — only the slotted content is `:global` now.

AlanBreck (approved):

PR #1437, opened 2026-07-30
Dialog: add configurable close button position

aguscruiz:
## Summary
- Adds a `closePosition` prop to Dialog (`'inline' | 'outside'`) that controls where the close button is rendered when `showClose` is enabled. Defaults to `'inline'`, so existing `showClose` usage is unchanged.
- **Inline** keeps the current behavior: close button inside the dialog (header when present, otherwise pinned at the top).
- **Outside** places the close button above the dialog, right-aligned to the top-right corner with an 8px (`--leo-spacing-m`) gap, using `--leo-color-icon-default` as the default icon/button color.
- Outside close stays a child of `<dialog>` (so it remains in the top layer for modals) with `overflow: visible`. Because the dialog background is transparent, header/body/actions get matching border radii so corners still look correct.
- **Responsive fallback:** when `closePosition="outside"` but the viewport is compact (below the dialog’s 480×480 breakpoint — `max-width: 479px` or `max-height: 479px`), the close button is forced back to the inline placement so it remains usable on small screens.
- Storybook: replaces the single “Show Close” story with “Show Close Inline” and “Show Close Outside” (open by default), and adds a `closePosition` control.

## Usage
```svelte
<!-- Existing behavior (default) -->
<Dialog showClose bind:isOpen>...</Dialog>

<!-- Explicit inline -->
<Dialog showClose closePosition="inline" bind:isOpen>...</Dialog>

<!-- Close outside the dialog (falls back to inline on compact viewports) -->
<Dialog showClose closePosition="outside" bind:isOpen>...</Dialog>
```

## Test plan
- [ ] Storybook → Dialog → **Show Close Inline**: close appears in the header; closing via the button works
- [ ] Storybook → Dialog → **Show Close Outside**: close sits above the top-right corner with an 8px gap; icon uses `color-icon-default`
- [ ] Outside variant: dialog corners still follow `--leo-dialog-border-radius` / `--leo-radius-xl`
- [ ] Outside variant: escape, backdrop click, and the close button all dismiss the dialog
- [ ] Outside variant on a compact viewport (< 480px wide or tall): close moves inside the dialog (inline)
- [ ] Resizing across the 480px breakpoint toggles outside ↔ inline without reopening the dialog
- [ ] `showClose` without `closePosition` still renders the inline close (backwards compatible)
- [ ] Inner Scroll story still shows the inline close correctly while scrolling

## Screens

image

image

## Responsive fallback - Move it inside the dialog when screen is small

image

### DIFF
--- a/src/components/dialog/dialog.stories.svelte
+++ b/src/components/dialog/dialog.stories.svelte
@@ -41,13 +41,20 @@
       'isOpen': {
         'control': 'none'
       },
+      'closePosition': {
+        'control': 'select',
+        'options': ['inline', 'outside'],
+        'description':
+          'Where to place the close button when showClose is enabled'
+      },
       'size': {
         'control': 'select',
         'options': ['mobile', 'normal']
       }
     },
     args: {
-      'isOpen': false
+      'isOpen': false,
+      'closePosition': 'inline'
     }
   }
 </script>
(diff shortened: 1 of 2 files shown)


aguscruiz:
I'm closing this PR as the two locations where it would make sense to have it no longer need it.

PR #1441, opened 2026-08-06
chore(socket-fix): migrate Socket CLI install to pnpm

kdenhartog:
The `socket fix` workflow currently installs the Socket CLI with `npm install -g` and only sets up Node.js. The Socket CLI itself requires `pnpm` to be on PATH (it fails with "Socket unable to locate pnpm"), so the workflow is broken.

This adds `pnpm/action-setup@v6.0.10` (pinned) so pnpm is available on the runner, and switches the global CLI install to `pnpm add -g @socketsecurity/cli`.

This workflow does not install project dependencies, so no `pnpm-lock.yaml` or `packageManager` changes are required.

### DIFF
--- a/.github/workflows/socket-fix.yml
+++ b/.github/workflows/socket-fix.yml
@@ -19,12 +19,18 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+      - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
         with:
-          node-version: '24.11.1'
+          run_install: false
+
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -97,5 +94,11 @@
     "leo-check": "./src/scripts/audit-tokens.js"
   },
   "license": "MIT",
-  "packageManager": "pnpm@11.1.2+sha512.415a1cc25974731e75455c1468371be74c5aa5fb7621b50d4056d222451609f11412f23fd602e6169f1e060466641f798597e1be961a10688836a67b16569499"
+  "overrides": {
+    "esbuild": "0.27.1",
+    "glob": "13.0.3",
+    "svgo@<2.8.2": "2.8.2",
+    "uuid": "14.0.0",
```
fallaciousreasoning: do we need this?

fallaciousreasoning (approved):
lgtm % confirming if the uuid change is needed

fallaciousreasoning (approved):

PR #1444, opened 2026-08-07
chore: use pnpm in socket-fix workflow

kdenhartog:


pnpm-lock.yaml
fallaciousreasoning, line 1: JS-yaml changes here seem like they should be reverted.

fallaciousreasoning (approved):

PR #1448, opened 2026-08-11
fix(socket-fix): author PR commits as brave-support-admin

kdenhartog:
## Problem

PR #1447 (generated by the `socket-fix` workflow) was authored by `github-actions[bot]` instead of `brave-support-admin`, while PR #1434 (generated by `update-icons`) is correctly authored by `brave-support-admin`. Both workflows use the same `peter-evans/create-pull-request` config (`author`, `committer`, `token`), so that isn't the difference.

## Root cause

`socket fix .` (with `CI: 'true'`) creates its **own git commit** using `SOCKET_CLI_GITHUB_TOKEN` before `peter-evans/create-pull-request` runs. The workflow was setting that to the default `GITHUB_TOKEN`, so GitHub attributes the commit to `github-actions[bot]`.

Evidence:
- PR #1447's head commit message is the Socket CLI's format (`fix: GHSA-… - <title>`), not the `commit-message: 'fix: address security advisories'` that `create-pull-request` would have written.
- The run log for `create-pull-request` shows `Configured git author as 'brave-support-admin <…>'` — but by then there were no uncommitted changes left to wrap in a new commit, so the PR inherited Socket's `github-actions[bot]` author.

`update-icons` doesn't hit this because nothing commits before `create-pull-request` runs.

## Fix

1. `SOCKET_CLI_GITHUB_TOKEN: ${{ secrets.LEO_UPDATE_ICONS_PAT }}` — same PAT `update-icons.yml` uses, so Socket's commit is authored by `brave-support-admin`.
2. Move the `Configure GPG` step **before** `Run Socket Fix` so `git config commit.gpgsign=true` is set when Socket makes its commit, making it GPG-verified.

## Verification

Next time the workflow is dispatched, the resulting PR's head commit should be authored by `brave-support-admin` and show as Verified.

### DIFF
--- a/.github/workflows/socket-fix.yml
+++ b/.github/workflows/socket-fix.yml
@@ -30,7 +30,10 @@ jobs:
           cache-dependency-path: '**/pnpm-lock.yaml'
 
       - name: Install Socket CLI
-        run: pnpm add @socketsecurity/cli@1.1.102
+        run: |
+          echo "${PNPM_HOME}/bin" >> "$GITHUB_PATH"
+          export PATH="${PNPM_HOME}/bin:$PATH"
+          pnpm add -g @socketsecurity/cli@1.1.102
 
       - name: Run Socket Fix
         env:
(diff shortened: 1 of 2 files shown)


kdenhartog:
Closing in favor of a clean rebase onto main.

PR #1459, opened 2026-08-19
[Dialog] Stop dialog content from widening the dialog

petemill:
`.leo-dialog` is `display: grid` but never set `grid-template-columns`, so content sat in an implicit `auto` column. An `auto` track can't be narrower than its items' min-content width, so content that can't break. For example, a `white-space: nowrap` + `text-overflow: ellipsis` title, grew the column past the dialog's own `max-width`. The dialog then scrolled horizontally (the top-layer UA style sets `overflow: auto`) and the content never truncated, because the row it lived in had become wider than the dialog. `min-width: 0` on the consumer's flex item doesn't help: it removes the floor on the item's used size, it doesn't cap what the item contributes to an ancestor's intrinsic width.

`grid-template-columns: minmax(0, 1fr)` drops the min-content floor but keeps the growth limit, so a definite-width dialog lays its content out at the declared width, while a dialog given an intrinsic `--leo-dialog-width` (`max-content`, `fit-content`) still hugs its content. Content that genuinely doesn't fit now overflows and scrolls at the dialog's declared width, rather than silently pushing the dialog past its `max-width`.

Adds a Truncating Content story: rows pairing unbreakable text with fixed-size actions, which is the case that used to blow the dialog out.

### Before
image

image

### After
image

image

### DIFF
--- a/src/components/dialog/dialog.svelte
+++ b/src/components/dialog/dialog.svelte
@@ -156,6 +156,15 @@
     margin: auto;
     border: none;
     display: grid;
+    /* The dialog sets its own width below, so content must never widen it. An
+     * implicit `auto` column can't be narrower than its items' min-content
+     * width, so unbreakable content (`white-space: nowrap` with
+     * `text-overflow: ellipsis`) would push the column past max-width and
+     * scroll the dialog horizontally instead of truncating. `minmax(0, 1fr)`
+     * drops that floor but keeps the growth limit, so an intrinsic
+     * `--leo-dialog-width` (`max-content`, `fit-content`) still hugs its
+     * content. */
+    grid-template-columns: minmax(0, 1fr);
     align-content: start;
 
     width: calc(100% - var(--leo-spacing-m) * 2);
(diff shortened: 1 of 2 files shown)


src/components/dialog/dialog.svelte
```diff
@@ -156,6 +156,15 @@
...
     border: none;
     display: grid;
+    /* The dialog sets its own width below, so content must never widen it. An
+     * implicit `auto` column can't be narrower than its items' min-content
+     * width, so unbreakable content (`white-space: nowrap` with
+     * `text-overflow: ellipsis`) would push the column past max-width and
+     * scroll the dialog horizontally instead of truncating. `minmax(0, 1fr)`
+     * drops that floor but keeps the growth limit, so an intrinsic
+     * `--leo-dialog-width` (`max-content`, `fit-content`) still hugs its
+     * content. */
```
AlanBreck, line 166: Nit: this feels verbose for what's needed.

AlanBreck (approved):
LGTM!

fallaciousreasoning (approved):

PR #1466, opened 2026-09-07
[Tokens]: Keep Figma composed colors dynamic across platforms

aguscruiz:
## Summary
- Preserve Figma composed colors as a referenced variable plus opacity, instead of baking the result into a static color.
- Emit CSS `color-mix()` with a hardcoded fallback inside `@supports`, Android ColorStateLists with `android:alpha` on API 23+, and Skia `ui::SetAlpha` recipes.
- Include the updated universal (and news) variable exports so composed tokens actually flow through the transformers.

## Test plan
- [ ] Run `pnpm run transform-tokens` and confirm composed tokens in `tokens/css/variables.css` have a hardcoded declaration plus a `color-mix()` override under `@supports`.
- [ ] Confirm Android composed colors land in `tokens/android/color*` (fallback) and `tokens/android/color*-v23` (dynamic alpha).
- [ ] Confirm Skia mixer output uses `ui::SetAlpha(...)` for composed tokens.
- [ ] Run `pnpm test`.


src/tokens/transformation/android/formats/colorStyle.js
```diff
@@ -14,7 +19,14 @@ const getTokenValue = (token) => {
 
 export default ({ dictionary, platform, options = {}, file }) => {
   const colorStyles = dictionary.allTokens
-    .filter((compositeToken) => compositeToken.type === 'color')
+    .filter(
```
fallaciousreasoning, line 23: I wonder if we should have some property like `function: 'COMPOSITE_COLOR'` in case there are more composites you want to support in the future

src/tokens/transformation/android/composedColors.js
```diff
@@ -0,0 +1,76 @@
+import { mkdirSync, rmSync, writeFileSync } from 'node:fs'
+import path from 'node:path'
+import { TinyColor } from '@ctrl/tinycolor'
+import { getTokenName } from './colorName'
+
+const outputDirectories = [
+  'color',
+  'color-night',
+  'color-v23',
+  'color-night-v23'
```
fallaciousreasoning: lets not add these separate v23 versions of the colors. Lets just make the new output the default

.tmp-composed-color-test.ts
fallaciousreasoning, line 1: this shoulnd't be committed

src/tokens/transformation/android/formats/colorStyle.js
aguscruiz: Good call. Figma already tags these with `expressionFunction: "COMPOSE_COLOR"`, so I switched detection to that instead of inferring from `referencedVariable` + `opacity`. There's a shared `isComposedColor` helper in `src/tokens/transformation/common/composedColor.ts` that all platforms use now, so another composite function later is just a new branch on that value.

src/tokens/transformation/android/composedColors.js
aguscruiz: Done — `color/` and `color-night/` now emit the dynamic ColorStateLists (`@color/<ref>` + `android:alpha`) as the default. The `*-v23` variants and the hardcoded fallbacks are gone.

.tmp-composed-color-test.ts
aguscruiz: Removed — that was a scratch harness that shouldn't have been in the commit.

fallaciousreasoning (approved):

